PR #50, opened 2015-01-30
Tests+small refactors

kuba-kubula:
Some changes were done to make the syntax more _"coffee-script"-like_

Also added a few tests for more rare configuration / runtime options


test/fixtures.coffee
```diff
@@ -874,22 +945,22 @@ sampleError = """
 """
 
 sampleHeaders = {
-"Content-Type": "application/json",
+"content-type": "application/json",
```
netmilk: This imho sucks. If you want to test some new case (introduce or verify some new funcionality), add fixtures and tests for that particular case and please do not change existing tests and fixtures .

I miss here some test like:
```
describe 'when expected and real headers keys are with different case', () ->
  it 'should not return any error', () ->
    ...
```

kuba-kubula: I'll change it back to it's original value. And I'll make a new test with the lower-cased + upper-cased versions then. Would that be ok?

netmilk: It would be excellent! Thanks a lot!

netmilk:
:candy:

PR #51, opened 2015-02-09
Json schema draft v4 support

netmilk:


scripts/build
```diff
@@ -8,5 +8,6 @@ if [ -d "src/" ]; then
 	#cp -a -R src/* lib/
 fi
 cp test/cucumber/features/node_js/Readme.md ./
-cp src/meta-schema.json lib/meta-schema.json
+cp src/meta-schema-v3.json lib/meta-schema-v3.json
+cp src/meta-schema-v4.json lib/meta-schema-v4.json
```
kuba-kubula: I'd prefer to use a `.coffee` file extension so the build script doesn't need to handle it by itself (coffee compile will handle it when building files). But YMMV.

src/mixins/validatable-http-message.coffee
```diff
@@ -14,11 +16,11 @@ class Validatable
   #@return [Object] :headers {ValidationErrors}, :body {ValidationErrors}, :statusCode [Boolean]
   validate: () ->
     @validation = {}
-
+    @validation.version = '2'
```
kuba-kubula: :+1:

src/mixins/validatable-http-message.coffee
```diff
@@ -14,11 +16,11 @@ class Validatable
...
     @lowercaseHeaders()
 
-    @validateHeaders() unless @headers == undefined
-    @validateBody() unless @body == undefined
+    @validateHeaders() unless @headers == undefined || @expected.headers == undefined
```
kuba-kubula: Why this change? It just brings another type of behavior. Do we really want "not to validate headers" if one of real/expected `headers` is missing?

dtto for `body`

src/mixins/validatable-http-message.coffee
```diff
@@ -14,11 +16,11 @@ class Validatable
...
+    @validateBody() unless @body == undefined || @expected.body == undefined
     @validateStatusCode() unless @statusCode == undefined
```
kuba-kubula: Why isnt' the same "or" as above used here too? (check for `expected.statusCode?`)

src/validators/json-schema.coffee
```diff
@@ -90,14 +127,25 @@ class JsonSchema
...
       indexes = [0..data.length - 1]
       indexes.forEach (index) ->
         item = data[index]
-        pointer = item['property'] or []
+
+        if item['property'] == undefined
+          pathArray = []
+        else if item['property'] == null
+          pathArray = []
+        else if item['property'] == [null]
```
kuba-kubula: You cannot do this type of comparison, it won't work in JS.
Please do `Array.isArray(item['property']) and item['property'].length is 1 and item['property'][0] is null`

Or simply we can remove these checks, they don't seem useable to me.
Can you please provide some way they can actually happen?

src/validators/json-schema.coffee
```diff
@@ -90,14 +127,25 @@ class JsonSchema
...
 
-    #amanda to gavel converter
+    #TV4 to gavel converter
     if data.length > 0
       indexes = [0..data.length - 1]
       indexes.forEach (index) ->
         item = data[index]
-        pointer = item['property'] or []
+
+        if item['property'] == undefined
```
kuba-kubula: This `if and the next `else if` can be written as `if not item['property]? then pathArray = []`

kuba-kubula:
Great thanks @marcelogo for the great work you've done to bring the basic `tv4`-based schema validations into _gavel.js_. I really appreciate the ground work.

yakovkhalinsky:
@marcelogo :+1:

PR #52, opened 2015-02-11
Highlight the code example in README

Baggz:


netmilk:
Nice, thanks!

kuba-kubula:
This change actually needs to be done in https://github.com/apiaryio/gavel/blob/master/node_js/Readme.md

@Baggz Would you please do that, so it is reflected in published npm package too? Thanks.

PR #53, opened 2015-02-24
Validation only content negotiation significant headers for value match

netmilk:
- Headers validated for match of value given in expected are:
```
[
  'content-type'
  'accept'
  'accept-charset'
  'accept-encoding'
  'accept-language'
]
```

- All other headers are validated only for their presence, so values is are validated.
- All dependencies updated to latest
- Better handling primitive types on root level in JSON example validation
  - Strings in JSONs on root level were causing exception
  - On the root level was validated only typo of array or object
  - All JSON primitive types are now always validated on JSON root level

### DIFF
--- a/src/validators/json-schema.coffee
+++ b/src/validators/json-schema.coffee
@@ -51,7 +51,7 @@ class JsonSchema
 
     if typeof @schema == 'string'
       try
-        @schema = JSON.parse(schema)
+        @schema = JSON.parse(@schema)
       catch error
         outError = new errors.SchemaNotJsonParsableError 'JSON validator: schema: ' + error.message
         outError['schema'] = @schema
--- a/test/cucumber/support/world.coffee
+++ b/test/cucumber/support/world.coffee
@@ -4,13 +4,13 @@ vm = require 'vm'
 util = require 'util'
 {assert} = require 'chai'
 
-# Function exported in this file is called before each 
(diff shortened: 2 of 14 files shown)


src/validators/json-schema.coffee
```diff
@@ -51,7 +51,7 @@ class JsonSchema
 
     if typeof @schema == 'string'
       try
-        @schema = JSON.parse(schema)
+        @schema = JSON.parse(@schema)
```
kuba-kubula: :+1:

netmilk:
Have a look at [IANA](http://www.iana.org/assignments/message-headers/message-headers.xhtml?utm_content=buffer0dda2&utm_medium=social&utm_source=twitter.com&utm_campaign=buffer)'s list of standard headers

test/cucumber/support/world.coffee
```diff
@@ -44,23 +44,23 @@ module.exports = () ->
...
       realOutputInspect = util.inspect(realOutput)
       expectedOutputInspect = util.inspect(expectedOutput)
-      
-      try 
+
+      try
         assert.deepEqual realOutput, expectedOutput
-      catch error 
+      catch error
         callback.fail "Output of code buffer does not equal. Expected output:\n" \
```
kuba-kubula: missing explicit `return`

PR #55, opened 2015-05-21
User friendly error messages

netmilk:
- get rid with confusing user facing messages with `null` media type
- get rid with redundant error messages from previous stages of validations
- custom messages for headers (get rid with JSON schema validator messages) 
- status code error message includes expected value

Note about implementation:

Tv4 validator can be supplied by custom validation messages, but the "At '/json/pointer'" part of the message is somewhere hardcoded and can't be customized. Due this limitation validation messages  from tv4 validator are parsed and sanitized to be human readable instead of some more sophisticated way how to do it.

### DIFF
--- a/src/mixins/validatable-http-message.coffee
+++ b/src/mixins/validatable-http-message.coffee
@@ -142,7 +142,8 @@ class Validatable
         @validation.body.realType = contentType
       catch error
         message = {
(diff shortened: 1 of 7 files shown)


kuba-kubula:
Awesome work. :heart:

PR #57, opened 2015-05-25
Showing expected header value in the error message

netmilk:


src/utils/tv4-to-headers-message.coffee
```diff
@@ -1,4 +1,6 @@
-module.exports = (message) ->
+caseless = require 'caseless'
+
+module.exports = (message, expectedHeaders) ->
   if message.indexOf("Missing required property:") > 1
```
kuba-kubula: should here be `> 1`? Not `> -1`?

netmilk: Probs yes ;)

kuba-kubula:
:100: points for using `caseless`.

netmilk:
🍬

PR #59, opened 2016-07-01
Updated dependencies

honzajavorek:
I had to pin Cucumber. Will resolve later, noted in https://github.com/apiaryio/gavel.js/issues/60.


package.json
```diff
@@ -11,33 +11,33 @@
   },
   "dependencies": {
     "amanda": "^0.5.1",
-    "async": "^1.0.0",
-    "caseless": "^0.10.0",
-    "clone": "^1.0.0",
-    "commander": "^2.6.0",
+    "async": "^2.0.0-rc.6",
```
miiila: Looks weird...

miiila:
Formal review done

package.json
honzajavorek: The thing is, this is RC and 2.X.X wasn't released yet, but I'd like gavel.js to work with it and pick it up as soon as it's released. So this is probably the only/best way to do it. Otherwise I'd have to `^v1.5.2` and that isn't what I want.

PR #61, opened 2016-07-18
Added browserfied version testing in phantom js to the build

netmilk:
@honzajavorek Please don't put me in shame and squash it when merging :)

PR #62, opened 2016-07-20
Test also by browserify

honzajavorek:
The jsonlint dependency taken from fork due to https://github.com/zaach/jsonlint/issues/56.
It's the same solution we are using internally in Apiary.

Supersedes https://github.com/apiaryio/gavel.js/pull/61. Inspired by https://github.com/apiaryio/gavel2html/pull/2.

### DIFF
--- a/scripts/mochify
+++ b/scripts/mochify
@@ -1,0 +1,3 @@
+#!/bin/sh
+cd $(dirname $0)/..
+./node_modules/.bin/mochify ./test/**/*.coffee --transform=coffeeify --extension=.coffee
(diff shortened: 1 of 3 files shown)


scripts/mochify
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+cd $(dirname $0)/..
+./node_modules/.bin/mochify ./test/**/*.coffee --transform=coffeeify --extension=.coffee
```
XVincentX: Just a note; please do not use shell scripts. They are not cross platform and super error prone.
Use node scripts and write javascript code if necessary.

PR #84, opened 2016-10-06
Remove jsonlint as dependency

### LINKED ISSUE #83: Install fails in restricted network environment (proxied npm registry)
Cannot install gavel (as a dependency of dredd) because the `jsonlint` dependency:
https://github.com/apiaryio/gavel.js/blob/master/package.json#L33

Please see this NPM post on how to ameliorate this issue:
http://blog.npmjs.org/post/145724408060/dealing-with-problematic-dependencies-in-a

> It’s usually a good idea to open an issue on the project with the Git dependency, politely asking the maintainers to replace the Git dependency, if possible. Generally, we discourage using Git dependencies in package.json, and it’s typically only used temporarily while a maintainer waits for an upstream fix to be applied and published.

lukateake:
Replaces jsonlint dependency with built-in JSON.parse to fix #83

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "http-string-parser": "0.0.5",
     "is-type": "0.0.1",
     "json-pointer": "^0.5.0",
-    "jsonlint": "josdejong/jsonlint",
     "media-typer": "^0.3.0",
     "tv4": "^1.2.7"
   },
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "is-type": "0.0.1",
     "json-pointer": "^0.5.0",
     "media-typer": "^0.3.0",
+    "parse-json": "^2.2.0",
     "tv4": "^1.2.7"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


honzajavorek:
@lukateake Thanks! This would solve #83, but I believe it degrades user experience since it's not possible to determine what's wrong with JSON from Gavel's output. And that is, I believe, the very reason why we use `jsonlint` here.

Did you try to find some alternative libraries on npm first? Maybe there are similar projects, which are alive and provide similar user experience.

lukateake:
@honzajavorek this change uses [parse-json](https://www.npmjs.com/package/parse-json) as a dependency which appears to be a well-maintained and widely used package from the prolific @sindresorhus. 

Best of all, it is truly "fit for purpose" in providing detailed information in the event of JSON errors.

honzajavorek:
I'm sorry for the delay, an autumn disease shoot me down for couple of days. Thanks very much for this! I'm quite excited to look at it next week.

honzajavorek:
I took a look at this and produced https://github.com/apiaryio/gavel.js/pull/85 as a successor. It seems there are some issues with browser tests.

PR #85, opened 2016-10-17
Replace jsonlint

honzajavorek:
The aim is to fix https://github.com/apiaryio/gavel.js/issues/83, i.e. to remove unmaintained `jsonlint` dependency while keeping the same functionality with detailed error messages if JSON parsing fails.

This PR is https://github.com/apiaryio/gavel.js/pull/84 successor. I took @lukateake commits, reworded them so they have Conventional Changelog prefixes and refactored a bit the error messages.
### Issues

When I made sure the detailed error message is tested in 9e2318d, I found out that while `parse-json` works correctly for node.js tests, it fails in browser tests:

```
expected

'Can\'t validate. Expected body \'Content-Type\' is \'application/hal+json\' but body is not a parseable JSON:\nundefined is not a constructor (evaluating \'Error.captureStackTrace(this, errorExError)\')'

to equal

'Can\'t validate. Expected body \'Content-Type\' is \'application/hal+json\' but body is not a parseable JSON:\nUnexpected token \'\\\'\' at 1:22\n{"creative?": false, \'creativ\': true }\n                     ^'
```

i.e. the parsing produces this error instead of the expected detailed error message:

```
undefined is not a constructor (evaluating \'Error.captureStackTrace(this, errorExError)
```

Unless this is resolved, I cannot get this merged. @lukateake @sindresorhus are you sure `parse-json` properly works in browser?

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "json-pointer": "^0.5.0",
     "jsonlint": "josdejong/jsonlint",
     "media-typer": "^0.3.0",
+    "parse-json": "^2.2.0",
     "tv4": "^1.2.7"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "http-string-parser": "0.0.5",
     "is-type": "0.0.1",
     "json-pointer": "^0.5.0",
-    "jsonlint": "josdejong/jsonlint",
     "media-typer": "^0.3.0",
     "parse-json": "^2.2.0",
     "tv4": "^1.2.7"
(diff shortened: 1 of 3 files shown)


honzajavorek:
@lukateake @alextes Unfortunately, it looks like `parse-json` isn't ready to be used in a browser out of the box - see https://github.com/sindresorhus/parse-json/issues/8. Gavel.js is being used also in the browser, so this is a clear blocker. I don't have a simple solution to this right now. Anyone is welcome to resume this effort with a proposal on how to solve this.

Issue #25, opened 2013-11-13
Gavel doesn't support $ref in schemas


ecordell:
Gavel uses Amanda for validating payloads against a JSON Schema. Unfortunately Amanda doesn't support `$ref` in schemas, and doesn't look like it will be added any time soon (see [this issue](https://github.com/Baggz/Amanda/issues/11) from two years ago).

This is a feature I would like to have in Gavel. It seems to me that the options are:
- Fork Amanda and add $ref support
- Use an alternative to Amanda for schema validation:
  - https://github.com/tdegrunt/jsonschema
  - https://github.com/akidee/schema.js
  - https://github.com/garycourt/JSV
  - https://github.com/zaggino/z-schema
  - Probably many more

joaosa:
@Almad any updates on this? :)

Almad:
@honzajavorek It's yours now :P

honzajavorek:
@joaosa I think Gavel should be able to currently validate both draft v3 (by Amanda?) and draft v4 (by tv4?), but to be honest, I have currently very little insight on what actually happens under the hood and would have to check.

Amanda has last commit 2 years ago and no new development should be expected there. Not sure what are @Baggz's plans with it.

As I said, I'd have to check, but

- if Gavel supports v4 and `$ref` works there, my best advice is that you should consider migrating to v4. 
- If `$ref` is not supported in Gavel's v4 implementation, then I would like us to do something about it. 
- If v4 is not supported in Gavel at all, then 😱 

My bet is that the second list item is true (again, this is not verified, it's from the top of my head). So if that's really true and you want to help us to get faster to a solution, you may want to contribute a failing test. Would that work for you?

joaosa:
@honzajavorek Thank you for the prompt response :)

I think I managed to produce two failing tests and added the corresponding schemas:

* added a test that should throw when invalid data is passed a valid $ref schema
    * the rationale is to ensure $ref schema validation works as expected.
    * https://github.com/joaosa/gavel.js/blob/master/test/unit/validators/json-schema-test.coffee#L193

* added a test that should throw when a schema with an invalid $ref is passed data
    * I added the data check as an attempt to ensure the data format wouldn't match the schema expectations, as according to [this](http://usingjsonschema.com/2014/08/04/why-isnt-my-ref-working.html), there might not be an error if the $ref URI can't be resolved (as `{}` will match anything).
[...]

PR #87, opened 2017-03-01
Detect errors on missing schema refs

joaosa:
More info [here](https://github.com/apiaryio/gavel.js/issues/25#issuecomment-283205325)

### DIFF
--- a/src/validators/json-schema.coffee
+++ b/src/validators/json-schema.coffee
@@ -171,12 +171,23 @@ class JsonSchema
   #@private
   validateSchemaV4: =>
     result = tv4.validateMultiple @data, @schema
+    validationErrors = result.errors.concat result.missing
 
     amandaCompatibleError =
-      length: result.errors.length
+      length: validationErrors.length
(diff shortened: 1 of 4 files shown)


src/validators/json-schema.coffee
```diff
@@ -171,12 +171,21 @@ class JsonSchema
...
+    validationErrors = result.errors.concat result.missing
 
     amandaCompatibleError =
-      length: result.errors.length
+      length: validationErrors.length
       errorMessages: {}
 
-    for error, index in result?.errors
+    for err, index in validationErrors
+      # need to create the error in case of missing schema errors
```
honzajavorek: I'd be even more verbose with the comment:

```coffee
# If tv4 meets $ref to a missing schema, it provides
# an "empty" error like this: ['', '': '']
```

joaosa: Updated the comment :)

honzajavorek:
Thanks! 👍

PR #93, opened 2017-06-01
Update package.json to pull jsonlint from npm

charleswisniewski:
This is related to this change:
https://github.com/zaach/jsonlint/pull/93


honzajavorek:
The browser tests don't pass with the current upstream jsonlint. Gavel.js needs to work in browser as well. Thanks very much for the effort, but I don't think the https://github.com/zaach/jsonlint/pull/93 will get merged (I commented there). I'm okay with re-opening this in case the situation changes.

PR #109, opened 2018-12-07
feat: replace jsonlint with json-parse-helpfulerror

abacaphiliac:
`jsonlint` is unmaintained, and the fork referenced in `gavel` is not published.

my proposal is based on the pull request from @honzajavorek where `jsonlint` is replaced with `json-parse-helpfulerror` instead of `parse-json`. `json-parse-helpfulerror` appears to work in the browser as well as server.

https://github.com/apiaryio/gavel.js/pull/85

### DIFF
--- a/src/mixins/validatable-http-message.coffee
+++ b/src/mixins/validatable-http-message.coffee
@@ -1,4 +1,4 @@
-jsonlint = require 'jsonlint'
+jph = require 'json-parse-helpfulerror'
 mediaTyper = require 'media-typer'
 validators = require '../validators'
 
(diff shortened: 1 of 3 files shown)


honzajavorek:
@abacaphiliac Wow! This would be awesome. BTW, credits for https://github.com/apiaryio/gavel.js/pull/85 belong to @lukateake https://github.com/apiaryio/gavel.js/pull/84 🙌 

I'll review this ASAP. If we get this merged and do the release train up to Dredd, we could finally close https://github.com/apiaryio/gavel.js/issues/83 and to remove [this whole section](https://dredd.readthedocs.io/en/latest/installation.html#installing-git) in Dredd's installation docs 🙏

src/mixins/validatable-http-message.coffee
```diff
@@ -138,19 +138,20 @@ class Validatable
...
+            but body is not a parseable JSON:\n#{error.message}\
+          """
           severity: 'error'
         }
-        message.message  = message.message + "\n" + error.message
         @validation.body.results.push message
     else
       try
-        JSON.parse @body
+        jph.parse @body
```
honzajavorek: In this case we're not handling the error, so I wonder whether using `JSON.parse()` wouldn't be faster 🤔 

OTOH, it could bring inconsistencies in how Gavel parsers JSON and what it considers to be valid or not - not sure whether there are any discrepancies, but [the probability is not low](http://seriot.ch/parsing_json.php). I guess we're better safe using the same implementation everywhere.

PR #138, opened 2019-05-02
Rewrites cucumber test definitions to JavaScript

artem-zakharchenko:
## Changelog

- Rewrites Cucumber test definitions in JavaScript
- Rewrites bash scripts in JavaScript
- Accounts for https://github.com/apiaryio/gavel-spec/pull/35
- Closes #137


scripts/cucumber.js
```diff
@@ -0,0 +1,41 @@
...
+const cucumber = spawn(
+  'node_modules/.bin/cucumber-js',
+  args.concat([
+    '-r',
+    'test/cucumber/support/',
+    '-r',
+    'test/cucumber/step_definitions/',
+    '-f',
+    'pretty',
+    '--compiler=coffee:coffee-script/register',
```
honzajavorek: is this line still needed?

artem-zakharchenko:
Experiencing flaky tests. Described here #139

scripts/cucumber.js
artem-zakharchenko: Nice catch. Should be removed.

Issue #96, opened 2017-11-21
Decaffeinate the codebase


honzajavorek:
The decaffeination allows for further development and removal of several CoffeeScript-specific dependencies. If the tests stay intact, the decaffeination can happen also by rewriting the code from scratch instead of using [decaffeinate](https://github.com/decaffeinate/decaffeinate):

1. support dual codebase, with both coffeescript and javascript
2. keep tests in coffeescript, rewrite code into javascript with the tests passing
3. once everything is in javascript and passing, rewrite tests to be in javascript
4. remove coffeescript from the project

-----

Related to https://github.com/apiaryio/dredd/issues/705 and the [Get Gavel.js back into shape](https://github.com/apiaryio/gavel.js/milestone/1) milestone.

PR #140, opened 2019-05-02
Rewrite Gavel to JavaScript

artem-zakharchenko:


honzajavorek:
Closes https://github.com/apiaryio/gavel.js/issues/96

PR #141, opened 2019-05-02
Clean dependencies

artem-zakharchenko:
## Changelog

- Removes CoffeeScript-related dependencies
- Adjusts the usage points of CoffeeScript dependencies
- Removes unnecessary dependencies (`is-type`)
- Removes dependencies that can be replaced by other existing dependencies (`lodash`)

### DIFF
--- a/lib/utils/schema-v4-generator.js
+++ b/lib/utils/schema-v4-generator.js
@@ -1,4 +1,3 @@
-const type = require('is-type');
 const getType = require('./get-type');
 
 const SCHEMA_VERSION = 'http://json-schema.org/draft-04/schema#';
@@ -31,7 +30,7 @@ class SchemaV4Generator {
    * @option {SchemaProperties} properties
    */
   constructor({ json, properties }) {
-    this.json = type.string(json) ? JSON.parse(json) : json;
+    this.json = typeof json === 'string' ? JSON.parse(json) : json;
 
     this.schema = undefined;
     this.properties = properties || new SchemaV4Properties({});
@@ -100,7 +99,9 @@ class SchemaV4Generator {
     const schemaType = this.getSchemaTypeFor(baseObject);
 
     const hasKeys =
-      type.object(baseObject) && Object.keys(baseObject).length > 0;
+      baseObject !== null &&
+      typeof baseObject === 'object' &&
+      Object.keys(baseObject).length > 0;
 
     if (schemaType === 'object') {
       if (properties.keysStrict) {
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "deep-equal": "1.0.1",
     "googlediff": "0.1.0",
     "http-string-parser": "0.0.6",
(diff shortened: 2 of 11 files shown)


artem-zakharchenko:
Test coverage is still removed, as it needs to be rewritten to use a different (perhaps, a more simple) solution than the current `scripts/cov`.

lib/utils/schema-v4-generator.js
```diff
@@ -100,7 +99,9 @@ class SchemaV4Generator {
     const schemaType = this.getSchemaTypeFor(baseObject);
 
     const hasKeys =
-      type.object(baseObject) && Object.keys(baseObject).length > 0;
+      baseObject !== null &&
```
honzajavorek: should this be `!==` or `!=`? (thinking of `undefined`)

honzajavorek: well `typeof undefined` is `'undefined'` so I guess it's sorted by the other line

package.json
```diff
@@ -30,24 +30,18 @@
...
     "http-string-parser": "0.0.6",
-    "is-type": "0.0.1",
     "json-parse-helpfulerror": "1.0.3",
     "json-pointer": "0.6.0",
     "media-typer": "0.3.0",
     "tv4": "1.3.0"
   },
   "devDependencies": {
     "chai": "4.2.0",
-    "coffee-coverage": "2.0.1",
```
honzajavorek: Closes https://github.com/apiaryio/gavel.js/pull/114

test/cucumber/support/world.js
```diff
@@ -193,7 +196,7 @@ class World {
 
   // Debugging helper
   inspect(data) {
-    if (typeof data === 'object') {
+    if (data !== null && typeof data === 'object') {
```
artem-zakharchenko: Added explicit comparison to `null` to prevent false positives (`typeof null` is `"object"`).

test/cucumber/support/world.js
```diff
@@ -5,6 +5,9 @@
  * Full docs: https://github.com/decaffeinate/decaffeinate/blob/master/docs/suggestions.md
  */
 const gavel = require('../../../lib/gavel');
+// DO NOT REMOVE!
+// lodash is not used, but somehow accessed from Cucumber feature testing.
+// May be a case of nasty shadow module references.
 const _ = require('lodash');
```
honzajavorek: how can the line work if we removed lodash from deps?

lib/utils/schema-v4-generator.js
artem-zakharchenko: First checking for `null` to prevent false positive on the next line. Yes, then the next line is satisfied only on `object`. `undefined` value would reject then.

test/cucumber/support/world.js
artem-zakharchenko: Okay, I'm testing now, and it seems to pass. Sorry for the confusing comment. Will remove.

PR #142, opened 2019-05-02
Update dependencies

honzajavorek:
Parked until @artem-zakharchenko drops Node 6 here. Closes https://github.com/apiaryio/gavel.js/issues/125.

When updating `media-typer`, I had to do something similar to this https://github.com/apiaryio/api-elements.js/pull/205. I did not upgrade Cucumber as it proved to be a rabbit hole - documented under https://github.com/apiaryio/gavel.js/pull/116.

### DIFF
--- a/lib/mixins/validatable-http-message.js
+++ b/lib/mixins/validatable-http-message.js
@@ -1,6 +1,6 @@
 const jph = require('json-parse-helpfulerror');
 const mediaTyper = require('media-typer');
-const contentTypeModule = require('content-type');
+const contentType = require('content-type');
 const isset = require('../utils/isset');
 const validators = require('../validators');
 
@@ -425,18 +425,21 @@ but body is not a parseable JSON:\n${error.message}\
     );
   }
 
-  isJsonContentType(contentType) {
-    if (!contentType) {
+  isJsonContentType(contentTypeValue) {
+    if (!contentTypeValue) {
       return false;
     }
     try {
-      const { type } = contentTypeModule.parse(`${contentType}`);
+      const { type } = contentType.parse(`${contentTypeValue}`);
       const parsed = mediaTyper.parse(type);
       return (
         (parsed.type === 'application' && parsed.subtype === 'json')
         || parsed.suffix === 'json'
       );
     } catch (e) {
+      // The Content-Type value is basically a user input, it can be any
+      // kind of rubbish, and it is neither this function's nor Gavel's problem
+      // if it's invalid
       return false;
     }
   }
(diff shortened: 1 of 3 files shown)


artem-zakharchenko:
Dropping of NodeJS version 6 covered in #143

honzajavorek:
@artem-zakharchenko Please look at the changes I made because of media-typer, whether they make sense to you. Thanks 🙏

lib/mixins/validatable-http-message.js
```diff
@@ -428,13 +429,16 @@ but body is not a parseable JSON:\n${error.message}\
...
-
-    return isJson;
+    try {
+      const { type } = contentTypeModule.parse(`${contentType}`);
+      const parsed = mediaTyper.parse(type);
+      return (
+        (parsed.type === 'application' && parsed.subtype === 'json')
+        || parsed.suffix === 'json'
+      );
+    } catch (e) {
```
artem-zakharchenko: We shouldn't silence an error. Can we log it to console?

lib/mixins/validatable-http-message.js
```diff
@@ -428,13 +429,16 @@ but body is not a parseable JSON:\n${error.message}\
...
     }
-
-    const parsed = mediaTyper.parse(`${contentType}`);
-    const isJson =
-      (parsed.type === 'application' && parsed.subtype === 'json') ||
-      parsed.suffix === 'json';
-
-    return isJson;
+    try {
+      const { type } = contentTypeModule.parse(`${contentType}`);
```
artem-zakharchenko: The only thing that troubles me is too many things sharing what seems the same namespace.
Perhaps, `contentType` argument is not really a proper content type? Or a `rawContentType`..?

lib/mixins/validatable-http-message.js
honzajavorek: The error gets thrown in case the value is not a valid media type. The only reponsibility of the function is to assess whether the value is JSON-related or not, so I think this is a valid approach.

The `contentType` value is basically a user input, it can be any kind of rubbish, and it is not Gavel's problem if it's invalid media type as far as it matches real/expected response.

lib/mixins/validatable-http-message.js
honzajavorek: Yeah, you're right basically I just hotfixed it. I'll try to make it nicer :)

honzajavorek: I renamed the variables in 519965b

lib/mixins/validatable-http-message.js
honzajavorek: I added an explanatory comment in 519965b so it doesn't trigger a red flag when someone reads it

PR #147, opened 2019-05-03
refactor: adds test coverage collector

### LINKED ISSUE #144: Rewrite test coverage collector
Rewrite existing test coverage collector script operating with CoffeeScript to the JavaScript version. Preferably, use native mocha output and coveralls anlyzis.

> Original coverage implementation commented in https://github.com/apiaryio/gavel.js/pull/132/commits/5e4441a2804d28919f511c83ade304046364d9b0

artem-zakharchenko:
Adds test coverage collector as a part of `ci:test`. Rewrites the way to get line coverage due to code base decaffeination.

## GitHub

- Closes #144

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,11 +12,12 @@
   "scripts": {
     "test": "npm run test:server && npm run test:browser && npm run test:features",
     "test:server": "mocha \"test/unit/**/*-test.js\"",
-    "test:server:coverage": "nyc --reporter=text-lcov npm run test:server | coveralls",
+    "test:server:coverage": "nyc npm run test:server",
     "test:browser": "mochify \"test/unit/**/*.js\"",
     "test:features": "node scripts/cucumber.js",
+    "coveralls": "nyc --reporter=text-lcov npm run test:server | coveralls",
     "ci:lint": "npm run lint",
-    "ci:test": "npm run test:server:coverage && npm run test:browser && npm run test:features",
+    "ci:test": "npm run coveralls && npm run test:browser && npm run test:features",
     "ci:release": "semantic-release"
   },
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -12,12 +12,11 @@
   "scripts": {
     "test": "npm run test:server && npm run test:browser && npm run test:features",
     "test:server": "mocha \"test/unit/**/*-test.js\"",
+    "test:server:coverage": "nyc --reporter=text-lcov npm run test:server | coveralls",
     "test:browser": "mochify \"test/unit/**/*.js\"",
     "test:features": "node scripts/cucumber.js",
-    "coverage": "scripts/cov",
-    "coveralls": "npm run coverage mocha-lcov-reporter | coveralls",
     "ci:lint": "npm run lint",
-    "ci:test": "npm run test",
+    "ci:test": "npm run test:server:coverage && npm run test:browser && npm run test:features",
     "ci:release": "semantic-release"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -14,10 +14,10 @@
     "test:server": "mocha \"test/unit/**/*-test.js\"",
     "test:browser": "mochify \"test/unit/**/*.js\"",
     "test:features": "node scripts/cucumber.js",
-    "coverage": "scripts/cov",
-    "coveralls": "npm run coverage mocha-lcov-reporter | coveralls",
+    "test:coverage": "nyc --reporter=text-lcov npm run test:server",
+    "coveralls": "npm run test:coverage | coveralls",
     "ci:lint": "npm run lint",
-    "ci:test": "npm run test",
+    "ci:test": "npm run test && npm run coveralls",
```
honzajavorek: This runs the tests twice, which will take the CI twice as much to finish the build. I think we don't really need the `coveralls` or `test:coverage` commands, maybe we could do something like

```
  "test:server": "mocha \"test/unit/**/*-test.js\""
  "test:server:coverage": "nyc --reporter=text-lcov npm run test:server"
  "ci:test": "npm run test:server:coverage (collect coverage) && npm run test:browser && npm run test:features && (send coverage to coveralls)"
```

The question is how to collect coverage like this. Does nyc support something like reporting and collecting coverage at the same time?

artem-zakharchenko: You are right, no need to run unit tests twice.
We need to keep in mind that the coverage is collected only from the unit (browser) tests. This can simplify a few things.

Within CI I don't see a problem to run test + collect coverage, since we are running tests anyway. I will integrate coverage collection without duplicate tests run.

package.json
```diff
@@ -12,12 +12,11 @@
...
     "test": "npm run test:server && npm run test:browser && npm run test:features",
     "test:server": "mocha \"test/unit/**/*-test.js\"",
+    "test:server:coverage": "nyc --reporter=text-lcov npm run test:server | coveralls",
     "test:browser": "mochify \"test/unit/**/*.js\"",
     "test:features": "node scripts/cucumber.js",
-    "test:coverage": "nyc --reporter=text-lcov npm run test:server",
-    "coveralls": "npm run test:coverage | coveralls",
     "ci:lint": "npm run lint",
-    "ci:test": "npm run test && npm run coveralls",
+    "ci:test": "npm run test:server:coverage && npm run test:browser && npm run test:features",
```
artem-zakharchenko: I duplicate tests pipeline for CI, substituting server test with test+coverage. That looks like a compromise between repeating code and introducing extra complexity. The goal is to execute the same pipeline on CI that you can on local.

Do you think this is a proper approach, @honzajavorek?

artem-zakharchenko:
I experience the same behavior when running `test:server:coverage` locally. I assumed that's because `coveralls` tries to send the coverage to some project (most likely Coveralls project), but fails to perform that on local. It does seem to pass during CI at the same time.

Completely agree on silent output issue. There should be an HTML reporter that would produce a coverage `.html` file, which we can then handle to `coveralls`. Perhaps that way we can still see the proper stdout _and_ have a coverage file as a result.

I know the test coverage is a tricky subject. I think it is still necessary, but one must not do coverage-driven development, as it rarely makes sense. For me coverage is useful in observing its delta between changes, which may indicate that a behavior of new things was not covered by tests.

honzajavorek:
> I assumed that's because coveralls tries to send the coverage to some project

Ah! That makes total sense, this didn't occur to me. I didn't realize there's `| coveralls` in the end so that could be it.

Issue #150, opened 2019-05-06
Refactor Gavel internals (non-breaking)


artem-zakharchenko:
Gavel internals can be refactored, preserving the same public API.

## Why

- To have a more readable code
- To have less code
- To remove classes and mutations

artem-zakharchenko:
Removing the milestone because the remaining changes are not breaking, and are entirely internal.

artem-zakharchenko:
Gavel validators partially refactored during #90. Now all validators are constructed with only expected data, and any actual data is accepted as an argument during the `validator.validate()` call.

There is still a room for improvement (i.e. obscure class extensions for HeadersExample), yet none of them are blocking.

PR #152, opened 2019-05-07
Adds ESLint

### LINKED ISSUE #146: Introduce ESLint+Prettier
Add linting and prettification of the code of GavelJS.

## Why

- To keep changes linted and prone to syntactic or logical errors
- Use Prettier to get rid of as much syntactic rules as possible

artem-zakharchenko:
## Changelog

- Adds `eslint` config and npm commands (using Dredd eslint config as an example)
- Adds linting to CI
- Fixes existing source code to prevent violation of eslint rules

## GitHub

- Closes #146

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,7 @@
+{
+  "useTabs": false,
+  "semi": true,
+  "trailingComma": "none",
+  "arrowParens": "always",
+  "singleQuote": true
+}
--- a/lib/mixins/validatable-http-message.js
+++ b/lib/mixins/validatable-http-message.js
@@ -135,7 +135,7 @@ class Validatable {
       this.validation.headers.realType === expectsJson &&
       this.validation.headers.expectedType === expectsJson
     ) {
-      return (this.validation.headers.validator = 'HeadersJsonExample');
+      this.validation.headers.validator = 'HeadersJsonExample';
     } else {
       this.validation.headers.validator = null;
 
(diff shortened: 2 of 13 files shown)


lib/mixins/validatable-http-message.js
```diff
@@ -275,22 +270,19 @@ not a parseable JSON:\n${error.message}\
...
-          };
-
-          this.validation.body.results.push(message);
+          });
         }
       } else {
         try {
           jph.parse(this.expected.body);
           this.validation.body.expectedType = 'application/json';
-        } catch (error3) {
```
honzajavorek: lol, error3

package.json
```diff
@@ -10,6 +10,7 @@
     "gavel": "bin/gavel"
   },
   "scripts": {
+    "lint": "eslint lib/**/*.js",
```
honzajavorek: not linting tests?

lib/mixins/validatable-http-message.js
artem-zakharchenko: Nasty variables shadowing after decaffeination. I've rewritten this block to be more sane, but more profound changes will be introduced under #150.

package.json
artem-zakharchenko: Good point. Added under 837df08.
I want to also add some linting in a watch mode, since the library doesn't have a compilation step, I'd like to prevent finding out about linting errors in CI. Needs to fail fast on local, while developing.

honzajavorek: 👍

.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "useTabs": false,
+  "semi": true,
+  "trailingComma": "none",
```
artem-zakharchenko: I do value trailing comma for git blame purposes, and I think it's also how Dredd is set up.
However, I've stumbled into issues when trailing commas in commonJS fail execution of a module in NodeJS.

honzajavorek: Dredd is set up to have trailing commas in objects/arrays, but not in functions. It was because of Node 6 support. Is is possible that the issues you've experienced were because of having trailing commas in function arguments and executing with Node 6?

https://github.com/apiaryio/dredd/blob/master/.eslintrc.js#L12:L20

PR #155, opened 2019-05-10
Modularize validation

### LINKED ISSUE #131: Charset header comparisons should be case insensitive.
See https://github.com/apiaryio/dredd/issues/117:

> I have a test on the correct charset being returned, but the test fails because the comparison is case sensitive (so 'UTF-8' is not equal to 'utf-8').

artem-zakharchenko:
These changes are meant to modularize the logic in Gavel and replace classes architecture by functions and functional composition.

## GitHub

- Relates to #150 
- Fixes #131 

## Roadmap

- [x] Validation units
  - [x] validateStatusCode
  - [x] validateHeaders
  - [x] validateBody
- [x] Public API
  - [x] validate
    - [x] validateElement
  - [x] ~~isValid~~ (out of scope)
  - [x] ~~isValidatable~~ (out of scope)

## Discoveries

- [x] Make the same call signature of validator units (right now some accept relevant data, but some the entire real/expected transaction elements)
- [x] Consider using `TextDiff` as a fallback body validator in case no matching validator is found
- [x] Consider exporting and testing `getBodyType` and similar utils to prevent deep testing hierarchy. Assert validity of the validation response on `validateBody`, assert validity of real/expected body types in the respective utility unit tests. (86304034498ae741289943a73140d438301316db)
- [x] Account for validation units like `validateBody` that need access to other fields of req/res (i.e. `headers`). It appears that normalization of transaction elements must happen before each individual validation unit is executed, meaning before invoking `validateReq`/`validateRes`.
- [x] Handle `headers` normalization. Ensure it's an Object, fallback to empty Object if it's not set, and throw exception if it's rubbish (aef5079bb4013c79e61d9024d59ff48eaf0011b3)
- [x] Align export of validation units (some have default exports, some named, it's shuffled) (72354d4)
- [x] Propagate the `error` arguments to the `callback` of `validate` (8ae92b4)
- [x] Adjust `getBodyType` error on non-parsable JSON to be different for real/expected content types (e8a6406377dc2a5e6c74b4e5f9bb911e8a091658)
- [x] Integrate new implementation into old unit tests (334f9ad)
- [x] Move normalization of real/expected away from `validate` (dc8a743)
- [x] Clean up old implementation and tests


lib/api/test/unit/validateBody.test.js
```diff
@@ -0,0 +1,110 @@
+const { assert } = require('chai');
+const validateBody = require('../../units/validateBody');
+
+describe.only('validateBody', () => {
```
honzajavorek: btw in case you rely on the CI passing under the PR, this runs only certain tests

honzajavorek: (FWIW, I used to commit .only() a lot to the point I wrote myself a git hook preventing me to commit if .only() was found in tests... but then I discovered I can ``npm test -- --grep=validateBody`` and I do that now most of the time)

artem-zakharchenko: Yes, true. Thanks for spotting out. It's a WIP committed, will be removed.

artem-zakharchenko:
Updated `validateBody` unit. I will highlight main changes below.

## Tuples

During various validation phases we often push parsing/other errors to the `results` (errors field of the validation result). To preserve that logic while not dragging error handling into independent utils I have introduces tuples of `[error, data]` shape. For example:

https://github.com/apiaryio/gavel.js/blob/d7495f00ee8006e3c6494cc8479a27b96b41af78/lib/api/units/validateBody.js#L45

`parseContentType` may fail trying to parse a given content type, but the responsibility of handling that error is delegated to the parental scope. Based on the notion, we can handle thrown errors of different nature as intended. For example, push them to the public `results`:

https://github.com/apiaryio/gavel.js/blob/d7495f00ee8006e3c6494cc8479a27b96b41af78/lib/api/units/validateBody.js#L186-L197

## Parsed body media type 

The main algorithmic difference introduced is the storage of body type as a parsed media type:

https://github.com/apiaryio/gavel.js/blob/d7495f00ee8006e3c6494cc8479a27b96b41af78/lib/api/units/validateBody.js#L64-L73

This way we can get a relevant body validator based on _already parsed_ body media type (previously we would have to parse body content type _again_):

https://github.com/apiaryio/gavel.js/blob/d7495f00ee8006e3c6494cc8479a27b96b41af78/lib/api/units/validateBody.js#L138-L154

Please, @honzajavorek, I would like to ask you to review `getBodyValidator` with extra cautious, since I've rewritten the logic, so I might have missed something. Let me know your opinion on the points above as well. Thanks!

lib/api/units/validateHeaders.js
```diff
@@ -0,0 +1,68 @@
+const { HeadersJsonExample } = require('../../validators/headers-json-example');
+
+const APIARY_JSON_HEADER_TYPE = 'application/vnd.apiary.http-headers+json';
+
+function normalizeHeaders(headers) {
```
honzajavorek: **Heads up:** One of the things we should do in the future is to support headers of the same name (e.g. [Set-Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie)). I don't think we should do it now, but just keep it in mind that one day in distant future, we might want change the input format for headers to `array` of `{ name, value }` instead of an `object` where names are keys.

lib/api/units/validateHeaders.js
```diff
@@ -0,0 +1,68 @@
...
+  return (
+    headers instanceof Object &&
```
honzajavorek: what happens if this is `false`? the function returns `false`?

lib/api/units/validateHeaders.js
honzajavorek: Also, there is this neat thingy https://github.com/request/caseless

lib/api/units/validateHeaders.js
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Validates given real and expected headers.
+ * @param {Object} realHeaders
+ * @param {Object} expectedHeaders
+ */
+function validateHeaders(realHeaders, expectedHeaders) {
+  const real = normalizeHeaders(realHeaders);
+  const expected = normalizeHeaders(expectedHeaders);
+  const realType = getHeadersType(real);
+  const expectedType = getHeadersType(expected);
```
honzajavorek: The logic is simple, the code looks verbose. We're getting two objects and we want to validate them. We know their types (two objects) and we know we want to validate them as headers. It's either an error or `application/vnd.apiary.http-headers+json`. I didn't read the code, but what else could Gavel possibly send to this function?

Perhaps we could make it much simpler with all the assumptions:

```js
try {
  const real = normalizeHeaders(realHeaders);
  const expected = normalizeHeaders(expectedHeaders);
} catch (e) {
  return ... // got some rubbish
}

const validator = new HeadersJsonExample(real, expected);
const rawData = validator.validate();

...

return {
  validator: 'HeadersJsonExample',
  realType: 'application/vnd.apiary.http-headers+json',
  expectedType: 'application/vnd.apiary.http-headers+json',
  rawData,
  results
};
```

lib/api/units/validateStatusCode.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Validates given real and expected status codes.
+ * @param {number} realStatusCode
+ * @param {number} expectedStatusCode
+ */
+function validateStatusCode(realStatusCode, expectedStatusCode) {
+  const validator = new TextDiff(
+    normalizeStatusCode(realStatusCode),
+    normalizeStatusCode(expectedStatusCode)
+  );
```
honzajavorek: **Sigh:** I wish this could be just something like `realStatusCode === expectedStatusCode` 😄 But let's keep all the `TextDiff` heavy lifting for now, this needs to be done step by step. What you did is right.

artem-zakharchenko: Agree. At first glance this can be done with a simple weak comparison.
Noted: https://github.com/apiaryio/gavel.js/issues/159

lib/api/units/validateBody.js
```diff
@@ -0,0 +1,233 @@
...
+
+const { TextDiff } = require('../../../lib/validators/text-diff');
+const { JsonExample } = require('../../../lib/validators/json-example');
+const { JsonSchema } = require('../../../lib/validators/json-schema');
+
+function isPlainText(mediaType) {
+  return mediaType.type === 'text' && mediaType.subtype === 'plain';
+}
+
+function isJson(mediaType) {
```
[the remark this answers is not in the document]
artem-zakharchenko: I'm keeping `mediaType` namespace because it's a valid type returned from `mediaTyper.parse()` function. If we used some type superset like TypeScript it would be a little more evident.

lib/api/units/validateBody.js
```diff
@@ -0,0 +1,233 @@
...
+      real.body,
+      usesJsonSchema ? expected.bodySchema : expected.body
+    );
+  const rawData = validator && validator.validate();
+  const validatorResults = validator ? validator.evaluateOutputToResults() : [];
+  results.push(...validatorResults);
+
+  return {
+    validator: ValidatorClass && ValidatorClass.name,
+    realType: mediaTyper.format(realType),
```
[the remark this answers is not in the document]
artem-zakharchenko: Vice versa: they are kept as parsed to prevent additional parsing, and formatted into a string in the validation result to retain the same output format.

lib/api/units/validateBody.js
```diff
@@ -0,0 +1,233 @@
...
+}
+
+/**
+ * Returns a tuple of error and body media type based
+ * on the given body and normalized headers.
+ * @param {string} body
+ * @param {Object} headers
+ * @returns {[error, bodyType]}
+ */
+function getBodyType(body, headers) {
```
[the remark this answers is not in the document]
artem-zakharchenko: This would complicate every place we call `getBodyType` to include safety check:

```js
getBodyType(..., real.headers && real.headers['content-type']
```

However, this can be handled during some sanitization of headers to make sure `headers` is always set to an empty object, when not present. Not sure how it would align with the existing Gavel's behavior, so I wouldn't introduce this change now.

Your suggestion is good, on the other hand.

lib/api/units/validateBody.js
```diff
@@ -0,0 +1,233 @@
...
+ */
+function parseContentType(contentType) {
+  try {
+    const { type } = contentTypeUtils.parse(`${contentType}`);
+    const mediaType = mediaTyper.parse(type);
+    return [null, mediaType];
+  } catch (error) {
+    return [error, null];
+  }
+}
```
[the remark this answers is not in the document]
artem-zakharchenko: My main concern is how broad `try/catch` can be applied as opposed to tuples.

I strongly dislike nesting of try/catch blocks, as it makes code unreadable:

https://github.com/apiaryio/gavel.js/blob/f58628f4cd2d837d6033f4abad99a3cd60d26b19/lib/mixins/validatable-http-message.js#L238-L289

Moreover, it makes error context matter, as if we throw in a wrong closure, a wrong `catch` handler will process it. Add it to combinable nature of try/catch blocks (children exception will be handled by a parent's `catch` block, if no closer `catch` block is met), and we also get high chances of bugs, with errors of one kind processed as if they are errors of the other.

Tuples, although verbose, do not introduce a separate closure for errors, and cannot be messed up with nesting. Also, my notion to have context-less errors is to change the validation output of Gavel in the future. I don't think it's correct to mix library-level warnings/errors (like "can't validate") with the actual validation errors (the diff). We can use a dedicated output collector for that, and there we would use errors in different context.

lib/api/units/validateBody.js
```diff
@@ -0,0 +1,233 @@
...
+
+  try {
+    jph.parse(body);
+    const [mediaTypeError, bodyMediaType] = parseContentType(
+      hasJsonContentType ? contentType : 'application/json'
+    );
+
+    if (mediaTypeError) {
+      throw new Error(mediaTypeError);
+    }
```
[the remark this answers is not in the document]
artem-zakharchenko: I think this `throw` here is not needed at all. At least, it doesn't align with the nature of the thrown error. Catch block below handled scenarios when `body` is not a parsable JSON. However, if `parseContentType` fails, this already means that the body has been parsed successfully. `mediaTypeError` should be handled separately (if it's even needed).

This is one more argument to not mix errors context, since in your example both failing `jph.parse` and `parseContentType` are handled with a single `catch` block, but bear different nature of errors. To provide respective handling for those errors one would introduce nesting of try/catch, and I'm somewhat against that (see the points above).

lib/api/units/validateHeaders.js
artem-zakharchenko: Thanks for insights. Didn't find a corresponding issue, created: https://github.com/apiaryio/gavel.js/issues/160

lib/api/units/validateHeaders.js
artem-zakharchenko: Yes. I guess it would be more obvious to return explicit `null` in that case. What do you think?

artem-zakharchenko: Returning `null` in e058feb4869ec53257a6f372e9e71b409fcc3985

lib/api/units/validateBody.js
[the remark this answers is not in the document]
artem-zakharchenko: Using a single `isJson` a6ba96180f7b12e941a47cccfcf0114f8a546c8a

lib/api/units/validateBody.js
```diff
@@ -0,0 +1,233 @@
...
+    // List JsonSchema first, because weak predicate of JsonExample
+    // would resolve on "application/schema+json" media type too.
+    [
+      JsonSchema,
+      (real, expected) => {
+        return (
+          isJson(real) &&
+          expected.type === 'application' &&
+          expected.subtype === 'schema' &&
+          expected.suffix === 'json'
```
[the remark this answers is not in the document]
artem-zakharchenko: Adding `isJsonSchema` cc2a52de5ff8c81e91a00a4686d49900709aa48b

lib/api/units/validateBody.js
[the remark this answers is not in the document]
artem-zakharchenko: `getBodyType` now accepts content-type header directly: 4a9853626c67eea8f350adb231cb0841f4dbbfdc

lib/api/units/validateBody.js
```diff
@@ -0,0 +1,233 @@
...
+      ? {
+          /**
+           * @TODO The same message for real/expected
+           * body assertion. Real/expected must be reflected
+           * in the error message.
+           */
+          message: `\
+Real body 'Content-Type' header is '${contentType}' \
+but body is not a parseable JSON:
+${parsingError.message}\
```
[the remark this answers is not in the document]
artem-zakharchenko: No, definitely not. Previously, Gavel operated on "raw" content type, which required to parse it multiple times on some occasions. To prevent that, I introduced the storage of content-type as parsed throughout the validation unit logic.

I completely agree, I've also voiced my concern in other comment that this `throw` statement on media type parsing error doesn't belong there. It's questionable if it even should be handled.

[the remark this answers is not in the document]
artem-zakharchenko: I wonder what defines `content type NOT parseable`?

Would it be legible for validation if you explicitly specify you expect a certain content type, but the real entity contains a different, or no content-type at all? I'm examining how Gavel works, and it uses weak comparison for headers, sometimes checking only the header names (without values), sometimes behaving like [`containsAllKeys`](https://www.chaijs.com/api/assert/#method_containsallkeys) from `chai`. I can see this leading to false positives.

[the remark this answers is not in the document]
honzajavorek: Yeah, reading the table again, I think perhaps the fact that content type isn't parseable is not even worth the warning in fact?

When I think about it, basically the user wants to know whether the real and expected HTTP messages fit, they probably don't care so much about whether the content type is parseable or not. Gavel doesn't validate the inputs according to HTTP specs, it just silently does a best guess and provides the verdict and the diff.

So I'd change "warns about invalid content type, returns application/json (for the purpose of validation)" to "returns application/json (for the purpose of validation)" in case the body is parseable as JSON, but the content type is rubbish. That way the exception gets silenced, as you suggest.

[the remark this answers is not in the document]
honzajavorek: One thing is validation of the body, where we can infer certain behavior based on the value of the content type header, or to use whatever effective media type we think is the best. In that case I think it's only our internal implementation detail that we're trying to parse the media type and that it can fail if the value is rubbish. I think we should silence the exception and return whatever we infer from the situation.

Another thing is validation of the headers. As far as I recall Gavel validates only presence of headers (i.e. keys), except of content negotiation headers (content type, accept, etc.), where it also looks at the values. When validating the content type header in the context of headers (not body) validation, we don't even need to parse it. It's plain 1:1 comparison. If it's specified it should be `application/json`, then `application/json; charset=utf-8` fails the validation. There is only one corner case I can see, that is `application/json; charset=utf-8` vs `application/json; charset=UTF-8`, which I think Gavel should consider the same - https://github.com/apiaryio/gavel.js/issues/131, but that is something we can fix as the next step. And still, it can be carried out gracefully (pseudocode):

```
function normalizeContentTypeValue(contentType) { // for the purpose of headers validation
  try {
    const mediaType = parseMediaType(contentType)
    if (mediaType.params.charset) {
      lowercase the charset, format the media type back to string, return
    }
    return contentType;
  } catch (e) {
    return contentType;
  }
}
```

lib/api/units/validateBody.js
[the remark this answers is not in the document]
honzajavorek: Yes, I'm sorry, I didn't formulate the comment well. I did see the flow, but wasn't sure why it is needed, i.e. why it would need to be parsed twice. Now I noticed the type is needed as a parameter for the `getBodyValidator(realType)` where it would need to be parsed twice and suddenly it all clicked 👍

lib/api/units/validateBody.js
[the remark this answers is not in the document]
honzajavorek: 👍 Yeah, I'd say we should make sure we can rely on the fact headers is always an object.

lib/api/units/validateBody.js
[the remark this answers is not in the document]
artem-zakharchenko: > Or maybe we can just lowercase the whole header value?

Absolutely. As you've said, it's a plain string comparison, and such should always be case-insensitive.

lib/api/units/validateHeaders.js
honzajavorek: I think the headers should come in as an object. If they're not, we should:

- coerce it to `{}` if we think it is a valid situation
- throw a runtime error if we think it is an invalid situation

I don't see a good reason why the rest of Gavel should call this function with anything else than an object, so I'm inclined to the second option. We should make sure there are clear interfaces with expectable types. If the surrounding code doesn't play along, let's change the surrounding code so it does. E.g. if Gavel takes whatever user passes into it and gives it to `validateHeaders()` as it is, perhaps it should rather make sure headers are either full of headers or default to an empty object. Or, if user passes rubbish, then Gavel should halt and crash with an error.

lib/api/units/validateBody.js
```diff
@@ -0,0 +1,233 @@
...
+        `,
+          severity: 'error'
+        }
```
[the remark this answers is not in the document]
artem-zakharchenko: Great observation. Changed under b24e69f.

lib/api/units/validateHeaders.js
artem-zakharchenko: But all the normalization/etc. functions are not expected to throw. The proper validator is assigned based on whether both header types are JSON. I'm not sure what try/catch would catch in this case.

honzajavorek: In that case it can be even simpler :) If we ensure the realHeaders and expectedHeaders are always {} on the input, then it's okay without any error handling

lib/api/units/validateBody.js
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  const [realTypeError, realType] = getBodyType(real.body, real.headers);
+  const [expectedTypeError, expectedType] = expected.bodySchema
+    ? getBodySchemaType(expected.bodySchema)
+    : getBodyType(expected.body, expected.headers);
+
+  if (realTypeError) {
+    results.push(realTypeError);
+  }
```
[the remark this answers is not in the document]
artem-zakharchenko: I would like to separate library-side warnings and actual validation errors in the future. The absence of validation due to a malformed data is not equal to the error during the validation.

This is something on my breaking refactoring list, which we can discuss later. What Gavel misses is the validation of the given data to be a valid representation of that data (hope I make sense here). That is to validate the given data if any req/res validation (diff) can even be done. The errors produced by data validation should be handled in their own way, and I agree on you regarding short circuiting on them.

So, to be clear, I never intended to handle data errors with tuples, only validation-related errors. It happens those are mixed now, so it gives such impression.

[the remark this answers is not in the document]
artem-zakharchenko: Yeah, there is no other severity level than "error" 😄 We can also redesign severity levels (or deprecate them) in the further sessions.

lib/api/units/validateBody.js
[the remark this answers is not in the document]
honzajavorek: > I strongly dislike nesting of try/catch blocks

I agree, we shouldn't do that.

> children exception will be handled by a parent's `catch` block

That's an unfortunate feature of how JavaScript works (you cannot filter by exception type in `catch`), yes, but I don't think it means we should throw the whole thing away because of it. It just means we need to be cautious when using the exceptions. Also generally, this is rather a feature than a bug. If you don't handle the exception, it crashes the software, which is good - that's fail-fast. So problem with parent blocks catching more than they should happens only if you're silencing errors. And one should be always very careful when doing that. Either it's expected, well contained behavior (if that errors, this defaults to XYZ), or it is top-level catch-all thing, which serializes the error to something user can read.

And there's always bug potential with error handling. E.g. if you forget to handle an error from a tuple, or a callback argument.

I'd say a good rule of thumb, at least in the code in this file, would be that we have small, dedicated helper functions, which throw errors. No nesting etc. The code using them should always be cautious to handle the error and decide what to do with it. In the example above I wrote it is well-contained, expected, even documented behavior, and there's a default value returned if the exception happens. IMHO if we keep using exceptions like that, it is a powerful tool.

> I don't think it's correct to mix library-level warnings/errors (like "can't validate") with the actual validation errors (the diff)

Agreed. But then it seems to me that all the helpers should use library-level errors and `validateBody()` has the responsibility to assess them and decide what does it mean for the user - either providing defaults, verdicts, or creating user-facing errors for the `results` array.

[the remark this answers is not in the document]
artem-zakharchenko: How would you recommend to handle `getBodyType` with try/catch?

When I'm doing:

```js
try {
  jph.parse(...) // attempt to parse, throw if not parsable, OK
  const mediaType = parseContentType(...) // this may throw too
} catch(error) {
 // handle non-parsable json
}
```

If `parseContentType` call throws, the logic would treat current body as non-parsable JSON, since the error will be inherited and processed by the underlying catch block.

lib/api/units/validateBody.js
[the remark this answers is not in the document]
artem-zakharchenko: Removed that throw block https://github.com/apiaryio/gavel.js/pull/155/commits/9c50d2f1216b5e1d47ed7d1eaaab28bf8db00fca.

lib/api/units/validateBody.js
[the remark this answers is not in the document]
artem-zakharchenko: Okay, I've reworked it under https://github.com/apiaryio/gavel.js/pull/155/commits/9c50d2f1216b5e1d47ed7d1eaaab28bf8db00fca.
I can see that it bypassed `null` values even previously, so in case of an error `parseContentType` can just return `null`. Seems like a reasonable compromise to me, since invalid content type is not an issue, it gets fallback to `application/json` exactly for that purpose.

In that block preceding `isJsonContentType` also calls `parseContentType`, so in case it fails, it would return `false` and the `application/json` fallback value would be used.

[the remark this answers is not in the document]
honzajavorek: My take would be something along the lines of:

```js
// Because "is this body parseable as JSON" is the main piece of information
// we're trying to get and convey in the mentioned code. Easy to unittest this way.
function isParseableJSON(body) {
  try {
    jph.parse(body);
    return true;
  } catch (error) {
    return false;
  }
}

...

// Implementation of the table (roughly) I drafted in the other comment, unittestable.
function getBodyType(contentType, bodyIsJSON) {
  if (!bodyIsJSON) return { type: 'text', subtype: 'plain' }; // effectively, if the content cannot be parsed as JSON, we're going to validate it as if it was plain text
  try {
    const mediaType = parseContentType(contentType);
    if (isJSON(mediaType)) return mediaType; // keeping application/hal+json and friends ... 🤔 is this actually useful for anything? the "effective" media type for validation is application/json
    return { type: 'application', subtype: 'json' }; // forcing the validation to process the body as if it was JSON
  } catch (error) {
    return bodyIsJSON
      ? { type: 'application', subtype: 'json' }
      : { type: 'text', subtype: 'plain' };
  }
}
```

Could be used as:

```js
const bodyType = getBodyType(headers['content-type'], isParseableJSON(body));
```

This is not a complete code, but you probably get the idea. Split to more contained unittestable functions, which do look nice even with exceptions in place. The surrounding code then only composes them together with occasional if and else. The reader doesn't have to understand how `isParseableJSON()` works, the name is enough.

One thing which occurred to me when assembling the code above is whether it actually helps with anything to treat the `+json` media types as special snowflakes. Or why is the decision so complicated? Perhaps the rule is as simple as "If Gavel can parse the body as JSON, it validates it as JSON. Otherwise it validates it as plain text."

I might be wrong, I didn't read all the code, but perhaps we don't have to fiddle with the content type at all if it doesn't effectively change anything?

```js
function validateBody(...) {
  ...

  if (isParseableJSON(body) {
    return ... // parse as if it was application/json
  } else {
    return ... // parse as if it was text/plain
  }
}
```

lib/api/units/validateBody.js
```diff
@@ -46,14 +45,14 @@ function parseContentType(contentType) {
...
 function isJsonContentType(contentType) {
-  // No need to explicitly check for "contentType" existence.
-  // In case given content type is invalid, "parseContentType"
-  // will return an error, which would be coerced to "false".
-  const [error, mediaType] = parseContentType(contentType);
-
-  // Silence an error on purporse because contentType
-  // may contain any kind of rubbish.
-  return error ? false : isJson(mediaType);
+  try {
```
honzajavorek: `parseContentType()` doesn't throw anymore, it returns `null` in case of problems, so this `try/catch` isn't necessary. `isJSON()` throws, because we feed it with `null` and it tries to do things like `.suffix` on `null`.

This could be cleaner:

```js
const mediaType = parseContentType(contentType);
return mediaType ? isJSON(mediaType) : false;
```

lib/api/units/validateHeaders.js
artem-zakharchenko: Good idea, but I think that there is a feature that asserts that error would be thrown. We can rework this together with the spec in the further changes.

lib/api/units/validateBody.js
artem-zakharchenko: Agree. Changed in 32b9f67.

honzajavorek:
I see the PR description mentions it fixes an issue, but none of the 30 commits has a `fix: ` prefix. I guess the one which implements the lowercasing should have it.

lib/api/test/unit/units/validateBody.test.js
```diff
@@ -0,0 +1,282 @@
...
+                'message',
+                `At '/bar' Missing required property: bar`
+              );
+            });
+          });
+        });
+      });
+
+      describe.skip('application/schema+json', () => {
+        // ...
```
honzajavorek: Missing?

lib/api/test/unit/units/validateBody/getBodySchemaType.test.js
```diff
@@ -0,0 +1,48 @@
+const { assert } = require('chai');
+const mediaTyper = require('media-typer');
+const { getBodySchemaType } = require('../../../../units/validateBody');
+
+describe('getBodySchemaType', () => {
+  describe('when given non-string schema', () => {
+    // ...
```
honzajavorek: Missing?

lib/api/test/unit/units/validateBody/getBodySchemaType.test.js
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  describe('when given json schema', () => {
```
honzajavorek:
```suggestion
  describe('when given JSON Schema', () => {
```

lib/api/test/unit/units/validateBody/getBodySchemaType.test.js
```diff
@@ -0,0 +1,48 @@
...
+        assert.isNull(res[0]);
+      });
+
+      it('returns "application/schema+json" media type', () => {
+        assert.deepEqual(res[1], mediaTyper.parse('application/schema+json'));
+      });
+    });
+  });
+
+  describe('when given non-json schema', () => {
```
honzajavorek:
```suggestion
  describe('when given non-JSON Schema', () => {
```

lib/api/test/unit/units/validateBody/getBodySchemaType.test.js
```diff
@@ -0,0 +1,48 @@
...
+    const res = getBodySchemaType('foo');
+
+    it('returns parsing error', () => {
+      assert.match(
+        res[0],
+        /^Can't validate: expected body JSON Schema is not a parseable JSON:/
+      );
+    });
+
+    it('returns no media type', () => [assert.isNull(res[1])]);
```
honzajavorek: Why is the assert in an array?

lib/api/test/unit/units/validateBody/getBodySchemaType.test.js
honzajavorek: Or even better, "when given non-JSON" (see my other comment)

lib/api/test/unit/units/validateBody/getBodySchemaType.test.js
```diff
@@ -0,0 +1,48 @@
...
+      it('returns no errors', () => {
+        assert.isNull(res[0]);
+      });
+
+      it('returns "application/schema+json" media type', () => {
+        assert.deepEqual(res[1], mediaTyper.parse('application/schema+json'));
+      });
+    });
+
+    describe('that contains array', () => {
```
honzajavorek: I think we could convey our intentions better. Now it seems objects and arrays are supported as JSON Schema. While that's technically true, I think we want to rather say that we don't check the contents, so anything will do as JSON Schema.

What about we merge the two tests ("that contains object", "that contains array") into one, which is titled "that contains JSON", and which will use `'[1, 2, 3]'` as the tested value so it is clear that anything, even invalid JSON Schema document, will pass this particular function just fine, that there are no additional checks, and that it is an expected behavior.

lib/api/test/unit/units/validateBody/getBodyType.test.js
```diff
@@ -0,0 +1,78 @@
+const { assert } = require('chai');
+const mediaTyper = require('media-typer');
+const { getBodyType } = require('../../../../units/validateBody');
+
+const jsonTypes = ['application/json', 'application/schema+json'];
+const nonJsonTypes = ['text/plain'];
+
+describe('getBodyType', () => {
+  describe('when given json-like content type', () => {
```
honzajavorek:
```suggestion
  describe('when given JSON-like content type', () => {
```

lib/api/test/unit/units/validateBody/getBodyType.test.js
```diff
@@ -0,0 +1,78 @@
...
+    jsonTypes.forEach((jsonType) => {
+      describe(`when given "${jsonType}" content type`, () => {
+        describe('and parsable json body', () => {
```
honzajavorek:
```suggestion
        describe('and parsable JSON body', () => {
```

lib/api/test/unit/units/validateBody/getBodyType.test.js
```diff
@@ -0,0 +1,78 @@
...
+          it('returns no errors', () => {
+            assert.isNull(res[0]);
+          });
+
+          it(`returns "${jsonType}" media type`, () => {
+            assert.deepEqual(res[1], mediaTyper.parse(jsonType));
+          });
+        });
+
+        describe('and non-parsable json body', () => {
```
honzajavorek:
```suggestion
        describe('and non-parsable JSON body', () => {
```

lib/api/test/unit/units/validateBody/getBodyType.test.js
```diff
@@ -0,0 +1,78 @@
...
+
+          it('uses "text/plain" as fallback media type', () => {
+            assert.deepEqual(res[1], mediaTyper.parse('text/plain'));
+          });
+        });
+      });
+    });
+  });
+
+  describe('when given non-json content type', () => {
```
honzajavorek:
```suggestion
  describe('when given non-JSON content type', () => {
```

lib/api/test/unit/units/validateBody/getBodyType.test.js
```diff
@@ -0,0 +1,78 @@
...
+    nonJsonTypes.forEach((nonJsonType) => {
+      describe(`when given "${nonJsonType}" content type`, () => {
+        describe('and parsable json body', () => {
```
honzajavorek:
```suggestion
        describe('and parsable JSON body', () => {
```

lib/api/test/unit/units/validateBody/getBodyType.test.js
```diff
@@ -0,0 +1,78 @@
...
+          it('returns no errors', () => {
+            assert.isNull(res[0]);
+          });
+
+          it('coerces to "application/json" media type', () => {
+            assert.deepEqual(res[1], mediaTyper.parse('application/json'));
+          });
+        });
+
+        describe('and a non-json body', () => {
```
honzajavorek:
```suggestion
        describe('and a non-JSON body', () => {
```

lib/api/test/unit/units/validateBody/getBodyValidator.test.js
```diff
@@ -0,0 +1,46 @@
+const { assert } = require('chai');
+const mediaTyper = require('media-typer');
+const { getBodyValidator } = require('../../../../units/validateBody');
+
+const getMediaTypes = (real, expected) => {
+  return [real, expected].map(mediaTyper.parse);
+};
+
+describe('getBodyValidator', () => {
+  describe('when given known media type', () => {
```
honzajavorek: what if given unknown media type?

lib/api/test/unit/units/validateBody/isJson.test.js
```diff
@@ -0,0 +1,29 @@
+const { assert } = require('chai');
+const mediaTyper = require('media-typer');
+const { isJson } = require('../../../../units/validateBody');
+
+describe('isJson', () => {
```
honzajavorek: Airbnb says this should be isJSON - https://github.com/airbnb/javascript#naming--Acronyms-and-Initialisms

lib/api/test/unit/units/validateBody/isJson.test.js
```diff
@@ -0,0 +1,29 @@
...
+  describe('returns true', () => {
+    it('when given valid json media type', () => {
```
honzajavorek:
```suggestion
    it('when given valid JSON media type', () => {
```

lib/api/test/unit/units/validateBody/isJson.test.js
```diff
@@ -0,0 +1,29 @@
...
+      assert.isTrue(isJson(mediaTyper.parse('application/json')));
+    });
+
+    it('when given json-compatible media type', () => {
```
honzajavorek:
```suggestion
    it('when given JSON-compatible media type', () => {
```

lib/api/test/unit/units/validateBody/isJson.test.js
```diff
@@ -0,0 +1,29 @@
...
+      assert.isTrue(isJson(mediaTyper.parse('application/schema+json')));
+    });
+  });
+
+  describe('returns false', () => {
+    it('when given non-json media type', () => {
```
honzajavorek:
```suggestion
    it('when given non-JSON media type', () => {
```

lib/api/test/unit/units/validateBody/isJsonContentType.test.js
```diff
@@ -0,0 +1,28 @@
+const { assert } = require('chai');
+const { isJsonContentType } = require('../../../../units/validateBody');
+
+describe('isJsonContentType', () => {
```
honzajavorek: ditto regarding letter casing

lib/api/test/unit/units/validateBody/isJsonSchema.test.js
```diff
@@ -0,0 +1,25 @@
...
+
+describe('isJsonSchema', () => {
+  describe('returns true', () => {
+    it('when given a json schema media type', () => {
+      assert.isTrue(isJsonSchema(mediaTyper.parse('application/schema+json')));
+    });
+  });
+
+  describe('returns false', () => {
+    it('when given json media type', () => {
```
honzajavorek:
```suggestion
    it('when given JSON media type', () => {
```

lib/api/test/unit/units/validateBody/isJsonSchema.test.js
```diff
@@ -0,0 +1,25 @@
+const { assert } = require('chai');
+const mediaTyper = require('media-typer');
+const { isJsonSchema } = require('../../../../units/validateBody');
+
+describe('isJsonSchema', () => {
+  describe('returns true', () => {
+    it('when given a json schema media type', () => {
```
honzajavorek:
```suggestion
    it('when given a JSON schema media type', () => {
```

lib/api/test/unit/units/validateBody/isJsonSchema.test.js
```diff
@@ -0,0 +1,25 @@
+const { assert } = require('chai');
+const mediaTyper = require('media-typer');
+const { isJsonSchema } = require('../../../../units/validateBody');
+
+describe('isJsonSchema', () => {
```
honzajavorek: ditto regarding letter casing

lib/api/test/unit/units/validateHeaders.test.js
```diff
@@ -0,0 +1,139 @@
...
+      },
+      {
+        headers: {
+          'content-type': 'application/json',
+          connection: 'keep-alive'
+        }
+      }
+    );
+
+    it('has proper validator', () => {
```
honzajavorek: I'd be specific: "has HeadersJsonExample validator" or "has application/vnd.apiary.http-headers+json as the real type"

Code sometimes changes and it's not clear what was the original intention with ambiguous words like "has proper this" or "has expected that" or "has correct this". Also, when running the tests, such description are not very informative in the terminal output.

lib/api/test/unit/units/validateHeaders.test.js
```diff
@@ -0,0 +1,139 @@
...
+    it('has proper expected type', () => {
+      assert.propertyVal(
+        res,
+        'expectedType',
+        'application/vnd.apiary.http-headers+json'
+      );
+    });
+
+    describe('has error', () => {
+      it('for each missing header', () => {
```
honzajavorek: The test says for each missing header, but it doesn't verify it. The test should have more than one missing header to verify there will be more than one results for each of them.

lib/api/test/unit/units/validateHeaders.test.js
```diff
@@ -0,0 +1,139 @@
...
+          `At '/content-type' Missing required property: content-type`
+        );
+      });
+
+      it('contains proper severity', () => {
+        assert.propertyVal(res.results[0], 'severity', 'error');
+      });
+    });
+  });
+
```
honzajavorek: MARK FOR MYSELF: THIS IS WHERE I PAUSED MY REVIEW (will continue later)

lib/api/test/unit/units/validateBody/isJson.test.js
artem-zakharchenko: I don't see them covering acronyms being _within_ a variable. They do allow to use acronyms in lowercase, but how about cammelCase then?

```
isJson = cammelCase(is + json)
```

honzajavorek: I think what they're trying to say is that `isJSON` is better than `isJson` although the later is correct according the camelCase definition. As I understand it, what they are trying to say is that `isHTTPtransaction` is better than `isHttpTransaction` as well, for readability reasons.

lib/api/test/unit/units/validateBody/getBodyValidator.test.js
artem-zakharchenko: As we discussed previously, `getBodyValidator` is designed to work supported media types only. Preceding `getBodyType` returns only supported media types.

If we want to make sure `getBodyValidator` can digest any kind of media types, then I can suggest to use `TextDiff` as a fallback. However, as of now there is no need to do that, as it's not exposed publicly and internally always operates on supported media types.

lib/api/test/unit/units/validateBody/getBodySchemaType.test.js
artem-zakharchenko: Typo. Will be fixed.

lib/api/test/integration/validateMessage.test.js
```diff
@@ -0,0 +1,361 @@
...
+
+    it('contains all validatable keys', () => {
+      assert.hasAllKeys(result, ['headers', 'body']);
+    });
+
+    describe('method', () => {
+      it.skip('compares methods');
+    });
+
+    describe.skip('headers', () => {
```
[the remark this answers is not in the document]
artem-zakharchenko: Leftovers, will be fixed.

lib/api/test/integration/validateMessage.test.js
```diff
@@ -0,0 +1,361 @@
...
+    it('returns validation result object', () => {
+      assert.isObject(result);
+    });
+
+    it('contains all validatable keys', () => {
+      assert.hasAllKeys(result, ['headers', 'body']);
+    });
+
+    describe('method', () => {
+      it.skip('compares methods');
```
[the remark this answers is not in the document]
artem-zakharchenko: Done in e260a4c5c2fbef655f95c35939580ff43c1ace08

lib/api/test/unit/units/validateBody/getBodyType.test.js
artem-zakharchenko: Done e260a4c5c2fbef655f95c35939580ff43c1ace08.

lib/api/test/unit/units/validateHeaders.test.js
artem-zakharchenko: Agree. Provided in 0cfab2767595f42f8caf112dc61a21b354baa33c

lib/api/test/unit/units/validateHeaders.test.js
artem-zakharchenko: Thanks for pointing out, some of the initial tests were left behind. I've updated all test suits to contain explicit intent.

lib/api/test/unit/units/validateBody/getBodySchemaType.test.js
artem-zakharchenko: I've changed the way it's asserted.

.eslintrc.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
     node: true
   },
   rules: {
+    'no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
```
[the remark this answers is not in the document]
artem-zakharchenko: It's used here currently: https://github.com/apiaryio/gavel.js/blob/e260a4c5c2fbef655f95c35939580ff43c1ace08/lib/api/units/validateBody.js#L140

We can always swap the order of arguments, but it feels as a wrong compromise imho. The order of arguments if a functional design decision, and alerts on unused variables is a developer-friendly reminder. I would prioritize the first one, but am always open to discussion :)

lib/api/units/normalize/normalizeHeaders.js
```diff
@@ -0,0 +1,35 @@
...
+  const headersType = typeof headers;
+
+  if (headersType === null || headersType !== 'object') {
+    throw new Error(
+      `Can't validate: expected "headers" to be an Object, but got: ${headersType}.`
+    );
+  }
+
+  return Object.keys(headers).reduce(
+    (acc, name) => ({
```
honzajavorek: `acc` or `normalizedHeaders`?

lib/api/units/validateHeaders.js
```diff
@@ -0,0 +1,67 @@
+const { HeadersJsonExample } = require('../../validators/headers-json-example');
+
+const APIARY_JSON_HEADER_TYPE = 'application/vnd.apiary.http-headers+json';
+
+function normalizeHeaders(headers) {
```
honzajavorek: Hmm, shouldn't we require from `normalize`? Seems like duplicate code.

lib/api/units/validateStatusCode.js
```diff
@@ -0,0 +1,30 @@
+const { TextDiff } = require('../../validators/text-diff');
+
+const APIARY_STATUS_CODE_TYPE = 'text/vnd.apiary.status-code';
+
+function normalizeStatusCode(statusCode) {
+  return String(statusCode).trim();
```
honzajavorek: Nitpick: I'm thinking, if the status code is undefined or null, it might be nicer to coerce it to empty string instead of `'null'` or `'undefined'`, so it looks better in the diff/error message. What about something like this?

```js
return statusCode == null ? '' : new String(statusCode).trim();
```

Do you think it's worth?

lib/api/test/unit/units/validateHeaders.test.js
```diff
@@ -123,17 +136,25 @@ describe('validateHeaders', () => {
...
+    describe('produces an error', () => {
+      it('has one error', () => {
+        assert.lengthOf(res.results, 1);
+      });
+
+      it('has "error" severity', () => {
+        assert.propertyVal(res.results[0], 'severity', 'error');
+      });
+
+      it('has exlpanatory message', () => {
```
honzajavorek: typo

lib/api/test/integration/validateMessage.test.js
```diff
@@ -90,47 +90,54 @@ describe('validateMessage', () => {
...
         assert.propertyVal(
           result.headers,
           'realType',
           'application/vnd.apiary.http-headers+json'
         );
       });
 
-      it('has no expected headers type', () => {
-        assert.propertyVal(result.headers, 'expectedType', null);
+      it('has "application/vnd.apiary.http-headers+json" expcted type', () => {
```
honzajavorek: typo: expcted

lib/api/test/unit/units/validateHeaders.test.js
honzajavorek: typo: exlpanatory

lib/api/test/integration/validateMessage.test.js
```diff
@@ -90,47 +90,54 @@ describe('validateMessage', () => {
...
+      //             result.headers.results[0],
+      //             'message',
+      //             `\
+      // No validator found for real data media type
+      // "application/vnd.apiary.http-headers+json"
+      // and expected data media type
+      // "null".`
+      //           );
+      //         });
+      //       });
```
honzajavorek: leftover?

lib/api/test/unit/units/validateBody/getBodyValidator.test.js
```diff
@@ -43,4 +43,24 @@ describe('getBodyValidator', () => {
...
+
+  //     describe(`${realContentType} + ${expectedContentType}`, () => {
+  //       const [error, validator] = getBodyValidator(real, expected);
+
+  //       it('...', () => {
+  //         console.log({ error, validator });
+  //       });
+  //     });
+  //   });
+  // });
```
honzajavorek: leftover?

lib/api/test/unit/units/validateBody/isJson.test.js
honzajavorek: I'm not strongly opinionated about this and I'm fine if we're not strict about it.

lib/api/test/unit/units/validateBody/getBodyValidator.test.js
honzajavorek: You're right, I'm sorry. I think I was misled by "when given known media type", which kinda implied there could be also the other case, so I pointed it out thinking about corner cases. Marking as resolved.

.eslintrc.js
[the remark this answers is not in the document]
honzajavorek: I'm fine with it, but perhaps it doesn't deserve a global rule if it's at a single place :)

lib/api/units/normalize/normalizeHeaders.js
```diff
@@ -0,0 +1,35 @@
...
+ * @returns {Object}
+ */
+const normalizeHeaders = (headers) => {
+  if (!headers) {
+    return {};
+  }
+
+  const headersType = typeof headers;
+
+  if (headersType === null || headersType !== 'object') {
```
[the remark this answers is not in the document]
artem-zakharchenko: Thanks for spotting this out. You are right, I think we should compare `headers == null` and then the type of the `headers`. Will update.

lib/api/units/validateHeaders.js
artem-zakharchenko: It's a leftover. Now the normalization of HTML messages happens in `validateMessage`, so all the validation units operate on already normalized data. I will remove this normalization at all.

lib/api/units/validateStatusCode.js
artem-zakharchenko: Definitely. Added this change and unit tests to cover `normalizeStatusCode`.

lib/api/test/integration/validateMessage.test.js
artem-zakharchenko: Seems to be an outdated file state.

Issue #159, opened 2019-05-14
Use weak comparison for statusCode validation


artem-zakharchenko:
I suggest to replace the usage of `TextDiff` for statusCode validation to weak comparison (`==`).

## Motivation

TextDiff brings some additional information irrelevant for statusCode (line/columns). Numbers comparison should be enough to provide meaningful validation results for status codes.

PR #168, opened 2019-05-21
Maps status code validation error to contain real/expected status codes

### LINKED ISSUE #167: Status code validation error message is wrong
Upon investigation of the [failing Dredd CI](https://travis-ci.org/apiaryio/dredd/jobs/534710787) after using the newest version of Gavel, I have found out that the error message for non-matching status code is not per spec.

## Current behavior

```
Real and expected data does not match.
```

> This is the default output of the `TextDiff` validator that Gavel uses for status codes validation.

## Expected behavior

```
Status code is '${real}' instead of '${expected}'
```

This is an explicit message that has been mapped manually previously:

https://github.com/apiaryio/gavel.js/blob/c03a0f917f5bb9ba37ea63e2171e06361c08e349/lib/mixins/validatable-http-message.js#L412-L420

It seems that this mapping has been lost during the refactoring. It needs to be added back for Gavel to behave per spec.

artem-zakharchenko:
Adds explicit mapping of `statusCode` validation results to produce expected error messages that include real and expected status codes. Fixes failing tests in Dredd pipeline when using the newest release of Gavel.

## GitHub

- Closes #167

### DIFF
--- a/lib/next/units/validateStatusCode.js
+++ b/lib/next/units/validateStatusCode.js
@@ -10,13 +10,23 @@ const APIARY_STATUS_CODE_TYPE = 'text/vnd.apiary.status-code';
 function validateStatusCode(real, expected) {
   const validator = new TextDiff(real.statusCode, expected.statusCode);
   const rawData = validator.validate();
+  const results = validator.evaluateOutputToResults();
 
   return {
     validator: 'TextDiff',
     realType: APIARY_STATUS_CODE_TYPE,
     expectedType: APIARY_STATUS_CODE_TYPE,
     rawData,
-    results: validator.evaluateOutputToResults()
+    results: results.map((result) =>
+      Object.assign({}, result, {
+        message:
+          result.message === 'Real and expected data does not match.'
+            ? `Status code is '${real.statusCode}' instead of '${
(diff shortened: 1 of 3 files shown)


lib/next/units/validateStatusCode.js
```diff
@@ -10,13 +10,23 @@ const APIARY_STATUS_CODE_TYPE = 'text/vnd.apiary.status-code';
...
   const rawData = validator.validate();
+  const results = validator.evaluateOutputToResults();
 
   return {
     validator: 'TextDiff',
     realType: APIARY_STATUS_CODE_TYPE,
     expectedType: APIARY_STATUS_CODE_TYPE,
     rawData,
-    results: validator.evaluateOutputToResults()
+    results: results.map((result) =>
```
artem-zakharchenko: It's safe to map since the output of `evaluateOutputToResults` is `[]` when no errors occurred.

artem-zakharchenko:
I've also updated the tests that didn't seem to assert this validation output previously.

artem-zakharchenko:
Absolutely agree. Will be covered in #159. Thanks!

PR #172, opened 2019-05-21
feat: uses plain strings comparison for statusCode validation

### LINKED ISSUE #159: Use weak comparison for statusCode validation

artem-zakharchenko:
Uses plain string comparison to validate `statusCode` of given HTTP messages. Uses strict comparison since statusCode is normalized to always be a string, and is coerced to an empty string when not set.

## GitHub

- Closes #159 

## Caveats

- One of the reasons we previously used `TextDiff` is because of the `rawData` is produces (result of `validator.validate()`). That `rawData` is asserted by Cucumber tests. We should watch out for that.

### DIFF
--- a/lib/next/units/validateStatusCode.js
+++ b/lib/next/units/validateStatusCode.js
@@ -1,5 +1,3 @@
-const { TextDiff } = require('../../validators/text-diff');
-
 const APIARY_STATUS_CODE_TYPE = 'text/vnd.apiary.status-code';
 
 /**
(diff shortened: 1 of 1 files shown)


honzajavorek:
How does the `rawData` look like? It's part of Gavel's public interface, so we cannot just omit it, but perhaps we can hardcode it if it can have just one or two values.

artem-zakharchenko:
I will see how it looks like depending on the validation result. It's an empty string when status codes are matching, but it's most likely an array-ish object (hello, amanda) when there were some errors from TextDiff.

artem-zakharchenko:
In case of invalid status code `TextDiff` used to produce a patch header containing text diff information:

```
rawData: '@@ -1,3 +1,3 @@\n-2\n+4\n 00\n'
```

While I don't see how this is useful in case of one-line strings comparison, I need to check the `gavel-spec` and existing tests to make sure nothing relies on this.

honzajavorek:
The thing is I'm afraid the Apiary frontend could rely on it when rendering the diffs. I'll take a look.

honzajavorek:
OK I think the rawData for status code isn't used anywhere, so we can ditch it. But it is still a breaking change.

artem-zakharchenko:
@honzajavorek Then I would change the merge destination to a dedicated branch. We may want to ship another breaking changes at once, or it doesn't matter?

honzajavorek:
It is okay. Just change the commit to something like this:

```
refactor: uses plain strings comparison for statusCode validation

Close #159 

BREAKING CHANGE: Gavel doesn't provide diff data for status code anymore
```

PR #174, opened 2019-05-22
feat: adds "isValid" property on the validation results

### LINKED ISSUE #166: Deprecate "isValid"
I suggest to deprecate `isValid` function from the Gavel's public API.

## Motivation

Under the hood `isValid` performs a full validation and trims its output to a single boolean value:

https://github.com/apiaryio/gavel.js/blob/577f9b57a54f6fee3a14c63b0bd1f50811f11d96/lib/mixins/validatable-http-message.js#L40-L61

It gives a false sense of optimization when in fact it's much more transparent to return the entire validation result to the end user.

## Suggestion

- Remove `isValid` from the public API
- Validation result must contain `valid` (`isValid`) property with the boolean value indicating the validity of the given HTTP message pairs:

```js
const { validate } = require('gavel')

const result = validate(realReq, expectedReq)
const { valid } = result
```

artem-zakharchenko:
## Changelog

- Adds `isValid` property on the validation result of Gavel
- Adjusts tests to assert a proper value of `isValid` depending on the validation's input
- Adds deprecation warning when using legacy `isValid()` function from Gavel

## GitHub

- Closes #166 

## Caveats

* Adding a new field to the public API may be a breaking change, considering it also means updating `gavel-spec` to include that field, which breaks previous Gavel spec implementations (https://github.com/apiaryio/gavel-spec/pull/42)

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -39,6 +39,16 @@ function getValidatableObject(real, expected, type) {
 }
 
 function isValid(real, expected, type, cb) {
+  console.warn(`\
+Usage of "isValid" is deprecated and will be removed in the future releases of Gavel.
+Use "isValid" property of the validation result instead:
+
+const { validate } = require('gavel');
+
+const result = validate(real, expected, 'request');
+const { isValid } = result;\
+`);
+
   return proxy(getValidatableObject(real, expected, type), 'isValid', cb);
 }
 
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -56,12 +56,7 @@ function isValidatable(real, expected, type, cb) {
   return proxy(getValidatableObject(real, expected, type), 'isValidatable', cb);
 }
 
-function validate(real, expected, type, cb) {
-  return proxy(getValidatableObject(real, expected, type), 'validate', cb);
-}
-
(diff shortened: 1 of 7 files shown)


artem-zakharchenko:
CI will not pass until `gavel-spec` has `isValid` in the validation result object.

artem-zakharchenko:
Gavel spec updated under https://github.com/apiaryio/gavel-spec/pull/42. Should be updated as a part of this pull request (won't pass on `master` due to missing new `isValid` property).

honzajavorek:
ok so close the PR from GK and upgrade it here

lib/validate.js
```diff
@@ -56,12 +56,7 @@ function isValidatable(real, expected, type, cb) {
   return proxy(getValidatableObject(real, expected, type), 'isValidatable', cb);
 }
 
-function validate(real, expected, type, cb) {
-  return proxy(getValidatableObject(real, expected, type), 'validate', cb);
-}
-
 module.exports = {
-  validate,
```
artem-zakharchenko: Removed legacy `validate` because we are already using `next/validate` since the last release.

PR #178, opened 2019-05-28
Implements v2

### LINKED ISSUE #68: Deprecate and kill async API
According to the [documentation](https://www.relishapp.com/apiary/gavel/v/0-1/docs/node-js/response-async-api), `gavel.js` offers an async API to perform validation and other operations.

According to the [code](https://github.com/apiaryio/gavel.js/blob/master/src/validate.coffee#L8), this is not an async API but a wrapper that is blocking the event loop anyway.

This is bad and harmful as people thinks that the validation is non blocking (so I do not get worried about giant payloads as is off process), while at the end of the day it is.

Please deprecate and kill this API as soon as possible.

### LINKED ISSUE #158: Add method validation for requests
I suggest to consider adding `method` field validation of requests.

## Why

At the moment we don't validate the `method` field (at least explicitly). Migrating toward more readable validation result Object that is being evolved with validation results of each field, it appears that explicit `method` validation would be beneficial.

## How

1. Add the `method` field on `validateRequest` function.
1. Use strings comparison for validation.

### LINKED ISSUE #163: Do not mix internal exceptions and validation results
Gavel currently stores internal exceptions (such as failed parsing) in the same key as public validation results - `results` key of the validation result object. I strongly suggest to re-consider this behavior, and stop mixing two kinds of errors in the Gavel's output.

## Why

1. It's invalid per exceptions' design.
1. It may lead to false testing results (such as https://github.com/apiaryio/gavel-spec/pull/39)
1. It may confuse validation results consumers, relying on validation errors, but getting internal ones

artem-zakharchenko:
These changes implement the next major version of Gavel per [defined spec](https://github.com/apiaryio/gavel-spec/pull/44). They must be merged once the spec lands.

## Refs

- Closes #68
- Closes #158 
- Closes #163 

## Changelog

- See [Gavel spec v2](https://github.com/apiaryio/gavel-spec/pull/44) to know the breaking and non-breaking changes

## Roadmap

- [x] Remove unused code (utils, tests, feature tests setups)
- [x] Decide on `uri` feature, implement if decided


artem-zakharchenko:
CI requires next `gavel-spec` release to pass successfully.

honzajavorek:
> CI requires next gavel-spec release to pass successfully.

I'm wondering whether it wouldn't help if we had those two in one repo 🤔

artem-zakharchenko:
@hj it's troublesome that spec leaves its own life, yet isn't that the original intent? So that there can be multiple implementation of gavel?

I would say we need to asses if that's useful, and how many Gavel implementation in other languages we have. If the effort is not worth a trouble, I would vote for moving `gavel-spec` into `gavel` package (or vice versa), and treating its test suites as behavior tests.

honzajavorek:
> yet wasn't that the original intent? So that there could be multiple implementations of gavel?

Yes, it was. I guess we could solve it for once when we have Dredd monorepo where all these things can live, together and separated at the same time.

> how many Gavel implementations in other languages we have

After 6 years of gavel.js existence we still have just gavel.js. The strategy did work out for dredd hooks, but not for gavel. It seems there is no audience to it, people want to use gavel as a core of other tools, not to use it separately. You can see the hopes were different - there's gavel-spec, and there's gavel CLI. I'm not sure how many people use the CLI, TBH, gavel as an end product has never took off.

artem-zakharchenko:
As of 71a02bcae529a6d01f0b44da88c996f5a5e4995e, gavel.js is compliant with the next version of `gavel-spec` (dcda152fb0f404d25aa90cf656be323502b57640).

## Roadmap

- [ ] Ensure `method` field is validated as per spec
- [ ] Ensure `uri` field is validated as per spec

bin/gavel
```diff
@@ -23,29 +24,24 @@ process.stdin.on('end', function() {
...
+    var expectedResponse = httpParser.parseResponse(expectedHttp.response);
 
     var requestResult = false;
     var responseResult = false;
 
-    gavel.isValid(realRequest, expectedRequest, 'request', function(
-      err,
-      result
-    ) {
+    gavel.validate(realRequest, expectedRequest, function(err, result) {
```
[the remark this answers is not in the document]
artem-zakharchenko: Indeed. Fixed in 00ac00a.

artem-zakharchenko:
Updated `data_model` steps in 7e48433.

Previously, code snippets described in features were compared to the parsed HTTP message (parsing done as a part of `world.js` cucumber setup, **not a part of the library's API**). Parsed HTTP message was wrapped in one of the following classes from the library's API:

* HttpRequest
* ExpectedHttpRequet
* HttpResponse
* ExpectedHttpResponse

On it own those classes did absolutely nothing to perform the parsing, or affect a provided HTTP message (I omit partial normalization during the validation because this is not what `data_model` feature asserts).

Since aforementioned classes have been removed from the library's public API, test setup cannot rely on dynamic class instantiation anymore:

```js
const foo = gavel[klass](data)
```

Also, since `gavel[klass]` doesn't affect the given payload, there is no need to use it during the assertion. I've removed dynamic class instantiation and compare _feature code snippet_ and _parsed HTTP message_.

**I do, however, find this test useless, as it doesn't involve testing actual library's API, and looks like test for the sake of test.** I would vote for removing it, which means removing `data_model.feature` from `gavel-spec` as well (not the library's feature and doesn't have to be in the spec).

artem-zakharchenko:
Updated to the latest release of `gavel-spec`, build passes. Need to ensure clean-up is done, and re-request a code review on this pull request.

honzajavorek:
Regarding https://github.com/apiaryio/gavel.js/pull/178#issuecomment-500736768, I think the only benefit left is documenting how over-the-wire HTTP messages map to Gavel input. If we have proper docs about how to run Gavel and how to structure its inputs, I think it is okay to remove the feature.

test/cucumber/support/world.js
```diff
@@ -172,6 +172,14 @@ class World {
     return parsed;
   }
 
+  // Hacky coercion function to parse expcected Boolean values
+  // from Gherkin feature suites.
+  toBoolean(string) {
+    if (string === 'true') return true;
+    if (string === 'false') return false;
+    return !!string;
+  }
```
honzajavorek: I think `JSON.parse('false')` produces false and `JSON.parse('true')` produces true. Perhaps we could use it instead, for a simplicity?

test/cucumber/step_definitions/validators_steps.js
```diff
@@ -26,20 +26,14 @@ module.exports = function() {
...
-        return this.validate((error, result) => {
-          if (error) {
-            callback(new Error(error));
-          }
-          this.booleanResult = result.isValid;
-          return callback();
-        });
-      });
+        this.booleanResult = result.isValid;
+        return callback();
```
honzajavorek: Note: try/catch around callback call will catch also errors happening in subsequent code (in the callback). Instead of fixing this I'd rather focus on upgrading cucumber in the repo though and simplifying the test steps as much as possible, with promises etc.

lib/units/validateURI.js
```diff
@@ -1,14 +1,60 @@
+const deepEqual = require('deep-equal');
+
 const APIARY_URI_TYPE = 'text/vnd.apiary.uri';
 
+/**
+ * Parses a given query string into an Object.
+ * @param {string} queryString
+ */
+const parseQueryString = (queryString) => {
```
honzajavorek: Why DIY and not https://nodejs.org/api/url.html ?

artem-zakharchenko:
I agree. Data models bear informative value, but they weren't asserting anything in the public's API. As discussed with @kylef, I also voted to include this models in the documentation for general knowledge, but remove from tests.

lib/units/validateURI.js
artem-zakharchenko: My biggest concern was that using `URL` class you cannot create relative urls:

```js
new URL('/admin')
// Exception: invalid URL
```

> There is a fair chance my unfamiliarity with the API gets in a way, so I'll learn more and conclude whether we can use native API for URL parsing.

It requires a `base`, which may mislead if present in the comparison diff.

There is a `url` module that can parse relative urls, but it's [marked as legacy](https://nodejs.org/api/url.html#url_legacy_url_api). The WHATWG URL API (mentioned above) is suggested to be used instead.

test/cucumber/support/world.js
artem-zakharchenko: How would this work if the `string` is `abc`?
Calling `JSON.parse('abc')` will throw, and if the `JSON.parse(string)` is positioned below the `!!string` it's rendered useless. Maybe I'm not fully understanding what you meant...

lib/units/validateURI.js
artem-zakharchenko: Using

```
const url = require('url')
``

replaces the previous functionality.

Comment (artem-zakharchenko) [reply lib/units/validateURI.js]: Changed in fcd0ccf.

**Comment by artem-zakharchenko** [comment]
I wonder, would `require('url')` resolve in the browser? As any dependency, I would expect `url` to be bundled in order to be executed, yet being a native nodejs dependency I'm not sure what would happen.

@honzajavorek what do you think about this?

File: lib/units/validateURI.js [hunk 5]
```diff
@@ -1,52 +1,34 @@
...
-  const hostname = parsed[1];
-  const queryString = parsed[3];
-
+  const { pathname, port, hash, query } = url.parse(uri, true);
   return {
-    hostname,
-    query: parseQueryString(queryString)
+    pathname,
+    port,
+    hash,
```
artem-zakharchenko: Accounting for previously missing `hash` in the URI.

test/cucumber/support/world.js
honzajavorek: I wanted to reply with more context, but then I got stuck not being able to find `toBoolean()` used. Where is it being called? I couldn't find `toCamelCase()` or `toPascalCase()` either.

honzajavorek:
I think https://github.com/apiaryio/gavel.js/pull/178#issuecomment-503014631 is a good point. I'm not sure about the consequences. If relative URLs are an issue my idea was to work around it with prefixing it with `example.com` just for the sake of parsing, but I see you found solutions already in fcd0ccf3d736d46048add8db9d2550c72e07f086

PR #180, opened 2019-05-28
fix: uses strict and weak coercion for real and expected messages

### LINKED ISSUE #179: Missing properties of HTTP message do not get normalized
## Current behavior

Normalization evolves a given HTTP message Object, meaning it will apply transformer functions only to the properties present in the original payload. In case the properties are missing, no normalization (and coercion) would happen.

## Expected behavior

* Keep normalization behavior as-is for `expected` HTTP message to prevent mutating a user's intentions
* Add `coerce` to the `real` HTTP message which would coerce potentially missing values to the ones that guarantee proper input for Gavel

artem-zakharchenko:
* Separates normalization and coercion to have a single coercion map used for both `real` and `expected` (but in different mode).
* Introduces strict and weak coercion modules.
* Uses strict coercion on `real` HTTP message to ensure even the missing data is coerced. This allows to have proper diffs with the expected message.
* Adds unit tests to `evolve` that supports `strict` mode now.
* Adds unit test for `validateMessage` that asserts expected headers, but given none

## GitHub

- Fixes #179

### DIFF
--- a/lib/next/test/unit/units/normalize/normalizeHeaders.test.js
+++ b/lib/next/test/unit/units/normalize/normalizeHeaders.test.js
@@ -5,10 +5,8 @@ const {
 
 describe('normalizeHeaders', () => {
   describe('when given nothing', () => {
-    const headers = normalizeHeaders(undefined);
-
-    it('coerces to empty object', () => {
-      assert.deepEqual(headers, {});
+    it('throws upon invalid headers value', () => {
+      assert.throw(() => normalizeHeaders(undefined));
     });
   });
 
--- a/lib/next/test/unit/utils/evolve.test.js
+++ b/lib/next/test/unit/utils/evolve.test.js
@@ -11,71 +11,108 @@ const unexpectedTypes = [
 ];
 
 describe('evolve', () => {
-  describe('evolves a given object', () => {
-    const res = evolve({
-      a: multiply(2),
-      c: () => null
-    })({
-      a: 2,
-      b: 'foo'
+  describe('weak mode', () => {
+    describe('evolves given object', () => {
+      const result = evolve({
+        a: multiply(2),
+        c: () => null
+      })({
+        a: 2,
+        b: 'foo'
+      });
+
+      it('returns object', () => {
+        assert.isObject(result);
+      });
+
+      it('evolves matching properties', () => {
+        assert.propertyVal(result, 'a', 4);
+      });
+
+      it('bypasses properties not in schema', () => {
+        assert.propertyVal(result, 'b', 'foo');
+      });
+
+      it('ignores properties not in data', () => {
+        assert.notProperty(result, 'c');
+      });
     });
 
-    it('returns object', () => {
-      assert.isObject(res);
+    describe('evolves given array', () => {
+      const result = evolve({
+        0: multiply(2),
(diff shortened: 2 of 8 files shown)


lib/next/test/unit/units/normalize/normalizeHeaders.test.js
```diff
@@ -5,10 +5,8 @@ const {
 
 describe('normalizeHeaders', () => {
   describe('when given nothing', () => {
-    const headers = normalizeHeaders(undefined);
-
-    it('coerces to empty object', () => {
-      assert.deepEqual(headers, {});
+    it('throws upon invalid headers value', () => {
+      assert.throw(() => normalizeHeaders(undefined));
```
artem-zakharchenko: Coercion is no longer a part of normalization so we must ensure we throw if `headers` is invalid value.

lib/next/test/unit/utils/evolve.test.js
```diff
@@ -11,71 +11,111 @@ const unexpectedTypes = [
...
+    });
   });
 
-  describe('throws when given unexpected data', () => {
-    unexpectedTypes.forEach(([typeName, dataType]) => {
-      it(`when given ${typeName}`, () => {
-        assert.throw(() => evolve({ a: () => null })(dataType));
+  /**
+   * Strict mode
+   */
```
honzajavorek: redundant comment? (the same information is already in the `describe` below)

lib/next/units/coerce/coerceHeaders.js
```diff
@@ -0,0 +1,8 @@
+// Coerces given headers to an empty Object in case not present.
+// Conceptually, diff between missing headers and empty headers
+// should be treated the same.
+const coerceHeaders = (headers) => {
+  return !headers ? {} : headers;
```
honzajavorek: can be simplified to `return headers || {}`

lib/next/utils/evolve.js
```diff
@@ -24,10 +24,37 @@ const evolve = (schema) => (data) => {
     );
   }
 
-  return Object.keys(data).reduce((acc, key) => {
+  // return Object.keys(schema).reduce((acc, expectedKey) => {
```
honzajavorek: leftover? 🚧

lib/next/utils/evolve.js
```diff
@@ -38,7 +65,19 @@ const evolve = (schema) => (data) => {
     /* eslint-enable no-nested-ternary */
 
     return isArray ? acc.concat(nextValue) : { ...acc, [key]: nextValue };
-  }, result);
+  };
+
+  const foo = Object.keys(data).reduce(reducer, result);
```
honzajavorek: foo?

lib/next/validateMessage.js
```diff
@@ -25,6 +35,7 @@ function validateMessage(realMessage, expectedMessage) {
     results.body = validateBody(real, expected);
   }
 
+  // Indicates the validity of a real message
```
honzajavorek:
```suggestion
  // Indicates the validity of the real message
```

artem-zakharchenko:
Thanks, @honzajavorek. Amendments provided. Please see once more. Thanks!

PR #182, opened 2019-05-29
Checks HTTP message properties on the "expected" message

### LINKED ISSUE #181: Determine validation necessity based on "expected" HTTP message properties
## Current behavior

Gavel checks if a property is set in "real" or in both "real" and "expected" to perform a validation of respective property.

## Expected behavior

Gavel performs validation of all properties specified in the "expected' HTTP message. If a user expects something, it must be validated. The absence of expected property is also a validation result that must be reported to the user.

artem-zakharchenko:
> BREAKING CHANGE: Validates all expected HTTP message properties

## Changes

- Checks expected HTTP messages properties to determine which properties to validate (previously: validated if a property is present in both expected and real messages)
- Adds tests that ensure _all_ expected HTTP message properties get validated, even if not present in the real message

## GitHub

- Closes #181 

## Roadmap

- [x] Adjust validation necessity logic
- [x] Add tests

### DIFF
--- a/lib/next/test/integration/validateMessage.test.js
+++ b/lib/next/test/integration/validateMessage.test.js
@@ -1,6 +1,26 @@
 const { assert } = require('chai');
 const { validateMessage } = require('../../validateMessage');
 
+const validator = (obj, expected) => {
+  it(`has "${expected}" validator`, () => {
+    assert.propertyVal(obj, 'validator', expected);
+  });
+};
+
+const createTypeAssertion = (typeName, propName) => (obj, expected) => {
+  it(`has "${expected}" ${typeName} type`, () => {
+    assert.propertyVal(obj, propName, expected);
+  });
+};
+
+const realType = createTypeAssertion('real', 'realType');
+const expectedType = createTypeAssertion('expected', 'expectedType');
+const noErrors = (obj) => {
+  it('has no errors', () => {
+    assert.lengthOf(obj.results, 0);
+  });
+};
+
 describe('validateMessage', () => {
   describe('with matching requests', () => {
     const request = {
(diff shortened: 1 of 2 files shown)


lib/next/validateMessage.js
```diff
@@ -14,24 +14,21 @@ function validateMessage(realMessage, expectedMessage) {
...
+  // However, we want to use the same coercion logic for any coercion type.
   const expected = normalize(coerceWeak(expectedMessage));
 
-  if (real.statusCode) {
+  if (expected.statusCode) {
     results.statusCode = validateStatusCode(real, expected);
   }
 
-  if (real.headers && expected.headers) {
+  if (expected.headers) {
```
artem-zakharchenko: I would use `evolve` for `expected` http message, but sometimes the predicate for keys present is different than the actual keys (i.e. it validates `body` when either `body` _or_ `bodySchema` are present). This would complicate the logic so I keep it simple.

lib/next/test/integration/validateMessage.test.js
```diff
@@ -461,4 +308,106 @@ describe('validateMessage', () => {
       });
     });
   });
+
+  describe('... validates all properties in expected message', () => {
```
honzajavorek: `...`?

artem-zakharchenko: Will rename, leftover.

artem-zakharchenko: I was unsure how to correctly name these test cases. They seem to be a sub-suites for non-matching req/res.

honzajavorek: What about "always validates properties which are expected" or something along those lines?

artem-zakharchenko:
@honzajavorek linked and checked on `dredd`—all tests are passing.

If you have some worries, could you please link me to an explicit test suite in Dredd?
It should work okay, as we have defined this behavior as expected. In case it fails, or behaves unexpectedly in Dredd, I would look into Dredd first. Perhaps, it inherited wrong Gavel's behavior from before.

honzajavorek:
I was 90% sure it shouldn't affect anything, because there's `isset(body)` so not set bodies won't get validated and empty bodies (empty strings) will. So if Dredd tests pass it's okay.

artem-zakharchenko:
@honzajavorek yeah, we explicitly check with `isset()` to prevent false positives for empty-string bodies :) This can be a good test suite for Gavel.

artem-zakharchenko:
I've created an issue to consider custom chai assertions: https://github.com/apiaryio/gavel.js/issues/184. Let's look at it separately. Thanks.

PR #183, opened 2019-05-30
Fixes wrong strings interpolation

artem-zakharchenko:
Minor changes focused on cleaning up test suites and fixing CoffeeScript migration bugs.

### DIFF
--- a/test/unit/mixins/validatable-http-message-test.js
+++ b/test/unit/mixins/validatable-http-message-test.js
@@ -1,4 +1,5 @@
-const assert = require('chai').assert;
+/* eslint-disable no-shadow */
+const { assert } = require('chai');
 const sinon = require('sinon');
 
 const { HttpResponse } = require('../../../lib/model/http-response');
@@ -31,7 +32,7 @@ describe('Http validatable mixin', () => {
     ];
 
     methods.forEach((method) => {
-      it('should have ' + method + ' method defined', () => {
+      it(`should have '${method}' method defined`, () => {
         assert.isFunction(HttpResponse.prototype[method]);
       });
     });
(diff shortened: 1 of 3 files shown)


test/unit/mixins/validatable-http-message-test.js
```diff
@@ -1193,12 +1192,12 @@ Unexpected token '\\'' at 1:22
...
-                  results.forEach((result) => {
-                    results.push(result.severity);
-                  });
-
-                  assert.include(results, 'error');
+                  const severities = instance.validation.body.results.map(
+                    (result) => {
+                      return result.severity;
+                    }
+                  );
```
honzajavorek: Might be nicer:

```js
const severities = instance.validation.body.results.map(result => result.severity);
```

test/unit/mixins/validatable-http-message-test.js
```diff
@@ -1299,7 +1298,8 @@ Unexpected token '\\'' at 1:22
 
             describe('when a validator throws an error', () => {
               before(() => {
-                invalidSchema = require('../../fixtures/invalid-schema-v4');
+                // eslint-disable-next-line global-require
+                const invalidSchema = require('../../fixtures/invalid-schema-v4');
```
honzajavorek: Why not to require this at the top of the file, like

```js
const INVALID_SCHEMA = require('../../fixtures/invalid-schema-v4');
```

artem-zakharchenko: To be honest, I have no idea. I think there was some intent to do it this way in the past, but I can't verify it.

test/unit/mixins/validatable-http-message-test.js
artem-zakharchenko: Good, will update it.

PR #199, opened 2019-06-19
feat: coerces missing "body" of the actual HTTP message

### LINKED ISSUE #198: Provide missing real body coercion
Missing `body` of the real HTTP message must be coerced to an empty string.

## Motivation

- Otherwise scenarios like
```
expected: "foo"
real: undefined
```
produce an internal exception instead of a validation result.

artem-zakharchenko:
- Closes #198

### DIFF
--- a/lib/units/validateBody.js
+++ b/lib/units/validateBody.js
@@ -201,9 +201,9 @@ function validateBody(expected, real) {
   if (validatorError) {
     // In case determined media types mismtach, check if the real is not missing.
     // Keep in mind the following scenarios:
-    // 1. Expcected '', and got '' (TextDiff/TextDiff, valid)
+    // 1. Expected '', and got '' (TextDiff/TextDiff, valid)
     // 2. Expected {...}, but got '' (Json/TextDiff, invalid, produces "missing real body" error)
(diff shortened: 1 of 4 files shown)


lib/units/validateBody.js
```diff
@@ -197,9 +199,22 @@ function validateBody(expected, real) {
...
   if (validatorError) {
-    errors.push({
-      message: validatorError
-    });
+    // In case determined media types mismtach, check if the real is not missing.
+    // Keep in mind the following scenarios:
+    // 1. Expcected '', and got '' (TextDiff/TextDiff, valid)
+    // 2. Expected {...}, but got '' (Json/TextDiff, invalid, produces "missing real body" error)
+    // 3. Expected {...}, but git "foo" (Json/TextDiff, invalid, produces types mismatch error).
+    if (expected.body !== '' && hasEmptyRealBody) {
```
artem-zakharchenko: This isn't the best solution in terms of cyclomatic complexity, but the one that works without introducing an entirely custom flow for missing body. This way we can always coerce missing actual "body" to an empty string and properly differentiate between three mentioned scenarios.

PR #200, opened 2019-06-20
fix: removes arrow functions to prevent context binding of Cucumber steps

artem-zakharchenko:
Provides adjustments for Cucumber step definitions to properly execute `gavel-spec` features.

Coupled with https://github.com/apiaryio/gavel-spec/pull/52.

### DIFF
--- a/test/cucumber/step_definitions/method_steps.js
+++ b/test/cucumber/step_definitions/method_steps.js
@@ -1,17 +1,17 @@
 module.exports = function() {
-  this.Given(
-    /^you expect HTTP message method "([^"]*)"$/,
-    (method, callback) => {
-      this.expected.method = method;
-      return callback();
-    }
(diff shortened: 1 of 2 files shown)


artem-zakharchenko:
Technical details for those who wonder: Cucumber steps definition has a context that is bind to the function that defines a step. Using arrow functions re-binds the context, so you lose the connection to the Cucumber environment (declared in `world.js`).

honzajavorek:
Also, the Cucumber in this repo is outdated, that's why it looks so hairy. You can see that newer versions of Cucumber [can have much simpler steps and setup](https://github.com/apiaryio/dredd-hooks-template/blob/master/features/support/steps.js), but we'd need to rewrite the existing ones here first and to upgrade Cucumber, and that's not an immediate priority.

Issue #139, opened 2019-05-02
Flaky test: Cucumber CLI scenarios


artem-zakharchenko:
During Cucumber feature testing there are two CLI scenarios that are flaky. Both seem to timeout randomly, and I cannot draw a clear conclusion what causes scenarios timeout.

```
1) Scenario: Read and validate real raw HTTP message from STDIN - node_modules/gavel-spec/features/cli.feature:18
   Step: Then exit status is 0 - node_modules/gavel-spec/features/cli.feature:24
   Step Definition: test/cucumber/step_definitions/cli_stepdefs.js:30
   Message:
     function timed out after 5000 milliseconds
2) Scenario: Exit status for invalid message - node_modules/gavel-spec/features/cli.feature:27
   Step: Then exit status is 1 - node_modules/gavel-spec/features/cli.feature:39
   Step Definition: test/cucumber/step_definitions/cli_stepdefs.js:30
   Message:
     function timed out after 5000 milliseconds
```

> See the [TravisCI job](https://travis-ci.org/apiaryio/gavel.js/jobs/527270297) 

Mentioned scenarios must be reviewed, rewritten (if necessary), and stabilized.

PR #204, opened 2019-06-21
test: introduces custom chai assertions

### LINKED ISSUE #184: Consider custom chai assertions
We seem to have a lot of repetitive assertion logic in tests. It may be beneficial to extend `chai`'s assertion to use common custom assertions more freely:

```js
assert.hasValidator(result.headers, 'JsonSchema')
assert.hasRealType(result.headers, 'text/plain')
```

artem-zakharchenko:
These changes introduce custom Chai assertions to use in Gavel unit and integration tests.

## Motivation

- To have less verbose test suites
- To have common assertions integrated into `chai` chainable operators. Common assertions include:
  - result/field validity
  - expected/real type
  - errors
- To get rid of implementation details in test, and centralize validation result references (to rely on validation result structure in a single place–custom assertions declaration)
- To enhance test exceptions with more information about the scenario (include stringified validation result and optional hints)

## GitHub

- Closes #184

## Roadmap

- [x] Implement custom assertions
  - [x] result and field validity
  - [x] validator
  - [x] expectedType
  - [x] realType
  - [x] errors / noErrors
- [x] Use `expect` instead of `assert` for better readability of custom assertions
- [x] Ensured consistency of assertions between test suites
- [x] ~~Consider using `to.not.have.validator` as a replacement for `.to.have.validator(null)`~~

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,9 +12,12 @@ module.exports = {
     'no-plusplus': 'off',
     'func-names': 'off',
 
+    // Disabled to allow "expect()" assertions with parameters:
+    // expect(foo).to.be.valid
+    'no-unused-expressions': 'off',
+
     // Temporary overrides. Logic to be rewritten.
     // TODO https://github.com/apiaryio/gavel.js/issues/150
-    'no-param-reassign': 'off',
-    'no-unused-expressions': 'off'
+    'no-param-reassign': 'off'
   }
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,6 +14,7 @@ module.exports = {
 
     // Temporary overrides. Logic to be rewritten.
     // TODO https://github.com/apiaryio/gavel.js/issues/150
-    'no-param-reassign': 'off'
+    'no-param-reassign': 'off',
+    'no-unused-expressions': 'off'
   }
 };
--- a/test/integration/validate.test.js
+++ b/test/integration/validate.test.js
@@ -1,29 +1,9 @@
-const { assert } = require('chai');
+const { expect } = require('../chai');
 const { validate } = require('../../lib/validate');
 
-const isValid = (result, expectedValid = true) => {
-  it(`sets "isValid" to ${JSON.stringify(expectedValid)}`, () => {
-    assert.propertyVal(result, 'isValid', expectedValid);
-  });
-};
-
-const validator = (obj, expected) => {
-  it(`has "${expected}" validator`, () => {
-    assert.propertyVal(obj, 'validator', expected);
-  });
-};
-
-const createTypeAssertion = (typeName, propName) => (obj, expected) => {
-  it(`has "${expected}" ${typeName} type`, () => {
-    assert.propertyVal(obj, propName, expected);
-  });
-};
-
-const realType = createTypeAssertion('real', 'realType');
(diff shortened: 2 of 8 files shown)


.eslintrc.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
 
     // Temporary overrides. Logic to be rewritten.
     // TODO https://github.com/apiaryio/gavel.js/issues/150
-    'no-param-reassign': 'off'
+    'no-param-reassign': 'off',
+    'no-unused-expressions': 'off'
```
artem-zakharchenko:
```js
expect(obj).to.be.valid
```
is considered a violation of this rule. Therefore, disabling the rule. Such violations are not present anywhere else in the code.

artem-zakharchenko:
Experienced #139, restarting the build.

.eslintrc.js
honzajavorek: I think this should be then a separate "section" if the ignores:

```js
... 

// Disabled to allow expect() expressions in tests
'no-unused-expressions': 'off',

// Temporary overrides. Logic to be rewritten
'no-param-reassign': 'off',
...
```

test/integration/validate.test.js
```diff
@@ -38,39 +18,44 @@ describe('validate', () => {
...
-      expectedType(result.fields.method, 'text/vnd.apiary.method');
-      realType(result.fields.method, 'text/vnd.apiary.method');
+      expect(result.fields.method).to.be.valid;
+      expect(result.fields.method).to.have.validator(null);
+      expect(result.fields.method).to.have.expectedType(
+        'text/vnd.apiary.method'
+      );
+      expect(result.fields.method).to.have.realType('text/vnd.apiary.method');
+
       noErrors(result.fields.method);
```
honzajavorek: `noErrors` is intentionally going to be a next step or was it forgotten?

test/unit/units/validateBody.test.js
```diff
@@ -395,12 +435,43 @@ describe('validateBody', () => {
...
+                `At '/firstName' Missing required property: firstName`
               );
             });
           });
         });
       });
     });
   });
+
+  // describe.only('when validating Json Schema V3 (Amanda)', () => {
```
honzajavorek: dramatic chipmunk

test/integration/validate.test.js
artem-zakharchenko: The next step. I haven't yet decided what would be a comfortable interface to assert Gavel's validation errors, so I left errors assertions behind.

test/unit/units/validateBody.test.js
artem-zakharchenko: Removed during rebase of the existing commit.

.eslintrc.js
artem-zakharchenko: Grouped in cbfb18c2dbb9856f00e7458dcd6f940d332168e0.

artem-zakharchenko:
I still would like to provide errors assertions as a part of this pull request, just to make it clear. Focusing on them now.

honzajavorek:
I think it is completely okay to merge this now and add them in a separate PR, but up to you.

artem-zakharchenko:
Custom assertions for the errors were also added, so the pull request should be complete!

I ditched an idea to have `to.not.have.validator`, because I think that conceptually means there is no `validator` property, which is not the expected behavior.

PR #212, opened 2019-06-25
Removes unused dependencies

### LINKED ISSUE #207: Clean up dependencies
Analyze and uninstall unused dependencies.

artem-zakharchenko:
This pull request reviews and uninstalls unused dependencies from `gavel`. No breaking changes are introduced.

## GitHub

- Closes #207

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -800,11 +800,6 @@
       "integrity": "sha512-+Ryf6g3BKoRc7jfp7ad8tM4TtMiaWvbF/1/sQcZPkkS7ag3D5nMBCe2UfOTONtAkaG0tO0ij3C5Lwmf1EiyjHg==",
       "dev": true
     },
-    "async": {
-      "version": "3.1.0",
-      "resolved": "https://registry.npmjs.org/async/-/async-3.1.0.tgz",
-      "integrity": "sha512-4vx/aaY6j/j3Lw3fbCHNWP0pPaTCew3F6F3hYyl/tHs/ndmV1q7NW9T5yuJ2XAGwdQrP+6Wu20x06U4APo/iQQ=="
-    },
     "async-each": {
       "version": "1.0.3",
       "resolved": "https://registry.npmjs.org/async-each/-/async-each-1.0.3.tgz",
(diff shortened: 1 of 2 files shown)


artem-zakharchenko:
# Dependencies analyzis

Including the insights about other dependencies that may be easy to remove.

## `clone`

Used only at `headers-json-example` validator to deep copy the original result from `JsonSchema` and mutate it with a new `message`:

https://github.com/apiaryio/gavel.js/blob/2eb036dab59440e2481bff615f5d127df989569a/lib/validators/headers-json-example.js#L74

## `caseless`

Used only in `tv4-to-headers-message` to reference `expectedHeaders` headers keys in a caseless way.

https://github.com/apiaryio/gavel.js/blob/2eb036dab59440e2481bff615f5d127df989569a/lib/utils/tv4-to-headers-message.js#L14

honzajavorek:
I thought commander is used for the CLI?

https://github.com/apiaryio/gavel.js/blob/0138366a7512a394c25570b28733f9782e752c8a/bin/gavel#L4

artem-zakharchenko:
Good catch.

That's the only place I found it being used:

https://github.com/apiaryio/gavel.js/blob/d79a0e1caac32f3fdc3a48a0f23553be4e36ed52/bin/gavel#L10

Does it only sets a version, or are there some side-effects in place that I'm not seeing?

honzajavorek:
> Does it only sets a version, or are there some side-effects in place that I'm not seeing?

That's out of scope of my memory. We'd need to dig into the ancient scrolls in the attic to learn about that.

artem-zakharchenko:
By examining the source code of `commander`, I can conclude that it's a util library to build CLI modules. It doesn't seem to do anything but setting the version of the program, as there is explicit syntax like `cmd.option()`,  `cmd.action()`, etc.

I'm not familiar with `commander`, but we don't see to use it at the moment to its full extent. At the same time, it's an approved dependency, so, perhaps, we should keep it?

honzajavorek:
So it looks like we don't need it 😄 But I think it is okay to keep it, not to add ourselves work. Up to you. I'm more concerned about the CI passing with non-executable Gavel CLI. I wonder if the CLI tests even run.

artem-zakharchenko:
`commander` module reverted and remains the dependency of `gavel` module. Should be ready to be merged.

PR #214, opened 2019-06-25
Adds json schema parsing check during body validation

### LINKED ISSUE #203: validateBody() throws in case it is provided with no schema
When my JSON Schema happened to be invalid JSON due to a typo, I got an error it is unparseable JSON. But Gavel didn't halt at that moment and continued to evaluate the situation further, which lead to an error being thrown:

```
error: TypeError: argument obj is required
 at Object.format (/Users/honza/Projects/apiary/node_modules/dredd/node_modules/media-typer/index.js:46:11)
 at validateBody (/Users/honza/Projects/apiary/node_modules/dredd/node_modules/gavel/lib/next/units/validateBody.js:225:30)
 at validateMessage (/Users/honza/Projects/apiary/node_modules/dredd/node_modules/gavel/lib/next/validateMessage.js:32:20)
...
```

That comes from

https://github.com/apiaryio/gavel.js/blob/master/lib/units/validateBody.js#L235

Which is caused by `expected.bodySchema` being evaluated to `undefined` on this line:

https://github.com/apiaryio/gavel.js/blob/master/lib/units/validateBody.js#L173

artem-zakharchenko:
Performs given JSON Schema parsing during body validation to prevent processing of invalid or malformed JSON Schema. **Note that this performs no actual validation either a JSON Schema is a valid JSON Schema.**

## GitHub

- Fixes #203 
- Relates to the issue's origin: https://github.com/apiaryio/drafter/issues/719
- Discovered originally in https://github.com/apiaryio/apiary/commit/6b6c1723153bb0e0e358a1e0cb94fd9dc1b9b277

### DIFF
--- a/lib/units/validateBody.js
+++ b/lib/units/validateBody.js
@@ -149,6 +149,27 @@ function getBodyValidator(realType, expectedType) {
 }
 
 /**
+ * Parses a given JSON Schema to ensure it's a valid JSON.
+ * @param {string} schema
+ */
+const validateJsonSchema = (schema) => {
+  try {
+    JSON.parse(schema);
+  } catch (error) {
+    throw new Error(`\
+Failed to parse a given JSON Schema:
+
+${schema}
+
+Please check if that is a valid JSON and a valid JSON Schema.
+
+Original parsing error:
+${error.stack}\
+    `);
+  }
+};
+
+/**
  * Validates given bodies of transaction elements.
  * @param {Object<string, any>} expected
  * @param {Object<string, any>} real
@@ -170,6 +191,13 @@ function validateBody(expected, real) {
     real.headers && real.headers['content-type'],
     'real'
   );
+
+  // Attempt to parse a given JSON Schema
+  // in case it's a raw string.
+  if (typeof expected.bodySchema === 'string') {
+    validateJsonSchema(expected.bodySchema);
+  }
+
   const [expectedTypeError, expectedType] = expected.bodySchema
     ? getBodySchemaType(expected.bodySchema)
     : getBodyType(
--- a/lib/units/validateBody.js
+++ b/lib/units/validateBody.js
@@ -102,10 +102,14 @@ function getBodySchemaType(bodySchema) {
   try {
     jph.parse(bodySchema);
     return [null, jsonSchemaType];
-  } catch (exception) {
-    const error = `Can't validate: expected body JSON Schema is not a parseable JSON:\n${exception.message}`;
+  } catch (error) {
+    // Gavel must throw when given malformed JSON Schema.
+    // See https://github.com/apiaryio/gavel.js/issues/203
+    throw new Error(`\
+Failed to validate HTTP message "body": given JSON Schema is not a valid JSON.
 
-    return [error, null];
+${error.message}\
+`);
   }
 }
 
@@ -149,27 +153,6 @@ function getBodyValidator(realType, expectedType) {
 }
 
 /**
- * Parses a given JSON Schema to ensure it's a valid JSON.
- * @param {string} schema
- */
-const validateJsonSchema = (schema) => {
-  try {
-    JSON.parse(schema);
-  } catch (error) {
-    throw new Error(`\
-Failed to parse a given JSON Schema:
-
-${schema}
-
-Please check that the given JSON Schema is a valid JSON.
-
-Original parsing error:
-${error.message}\
-    `);
-  }
-};
-
-/**
  * Validates given bodies of transaction elements.
  * @param {Object<string, any>} expected
  * @param {Object<string, any>} real
@@ -191,14 +174,6 @@ function validateBody(expected, real) {
     real.headers && real.headers['content-type'],
     'real'
   );
-
-  // Attempt to parse a given JSON Schema
-  // to verify that it's a valid JSON.
-  // TODO Move this into an input validation layer
-  // once such is introduced.
-  if (typeof expected.bodySchema === 'string') {
-    validateJsonSchema(expected.bodySchema);
-  }
 
   const [expectedTypeError, expectedType] = expected.bodySchema
     ? getBodySchemaType(expected.bodySchema)
--- a/lib/units/validateBody.js
+++ b/lib/units/validateBody.js
@@ -161,10 +161,10 @@ Failed to parse a given JSON Schema:
 
 ${schema}
 
-Please check if that is a valid JSON and a valid JSON Schema.
+Please check that the given JSON Schema is a valid JSON.
 
 Original parsing error:
-${error.stack}\
+${error.message}\
     `);
   }
 };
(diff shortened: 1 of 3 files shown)


test/unit/units/validateBody.test.js
```diff
@@ -444,4 +444,32 @@ describe('validateBody', () => {
...
+          }
+          `
+        },
+        {
+          body: '{ "foo": "bar" }'
+        }
+      );
+
+    it('must throw with malformed JSON schema error', () => {
+      expect(getResult).to.throw(/^Failed to parse a given JSON Schema:/g);
```
artem-zakharchenko: Explicitly throw an exception and stop the processing.

lib/units/validateBody.js
```diff
@@ -148,6 +148,27 @@ function getBodyValidator(realType, expectedType) {
...
+  } catch (error) {
+    throw new Error(`\
+Failed to parse a given JSON Schema:
+
+${schema}
+
+Please check if that is a valid JSON and a valid JSON Schema.
+
+Original parsing error:
+${error.stack}\
```
kylef: If I understand correctly, this message would make it to the user, perhaps it isn't a good idea to show them our internal stack trace.

artem-zakharchenko: I didn't realize this would include our stack trace as well. Do you think removing stack trace for good would be a viable option here?

lib/units/validateBody.js
```diff
@@ -148,6 +148,27 @@ function getBodyValidator(realType, expectedType) {
...
+const validateJsonSchema = (schema) => {
+  try {
+    JSON.parse(schema);
+  } catch (error) {
+    throw new Error(`\
+Failed to parse a given JSON Schema:
+
+${schema}
+
+Please check if that is a valid JSON and a valid JSON Schema.
```
artem-zakharchenko: I would rephrase this error, as there is no JSON Schema validation in place.

It should say something like
```
Please check that the given JSON Scheme is a valid JSON.
```

lib/units/validateBody.js
kylef: I think `error.stack` will include the full stack, perhaps in the case of the traffic inspector or a hosted closed source product, the stack trace could contain implementation details outside gavel.

I am not sure the stack will give the user much help in fixing their problem, but `error.message` probably will, it should contain the underlying cause and source map information.

For example:

```
> let error;
> try { JSON.parse('{"href"}') } catch (e) { error = e }
SyntaxError: Unexpected token } in JSON at position 7
    at JSON.parse (<anonymous>)
    at repl:1:12
    at Script.runInThisContext (vm.js:96:20)
    at REPLServer.defaultEval (repl.js:332:29)
    at bound (domain.js:395:14)
    at REPLServer.runBound [as eval] (domain.js:408:12)
    at REPLServer.onLine (repl.js:639:10)
    at REPLServer.emit (events.js:194:15)
    at REPLServer.EventEmitter.emit (domain.js:441:20)
    at REPLServer.Interface._onLine (readline.js:290:10)

> error
SyntaxError: Unexpected token } in JSON at position 7
    at JSON.parse (<anonymous>)
    at repl:1:12
    at Script.runInThisContext (vm.js:96:20)
    at REPLServer.defaultEval (repl.js:332:29)
    at bound (domain.js:395:14)
    at REPLServer.runBound [as eval] (domain.js:408:12)
    at REPLServer.onLine (repl.js:639:10)
    at REPLServer.emit (events.js:194:15)
    at REPLServer.EventEmitter.emit (domain.js:441:20)
    at REPLServer.Interface._onLine (readline.js:290:10)

> error.message
'Unexpected token } in JSON at position 7'

> error.stack
'SyntaxError: Unexpected token } in JSON at position 7\n    at JSON.parse (<anonymous>)\n    at repl:1:12\n    at Script.runInThisContext (vm.js:96:20)\n    at REPLServer.defaultEval (repl.js:332:29)\n    at bound (domain.js:395:14)\n    at REPLServer.runBound [as eval] (domain.js:408:12)\n    at REPLServer.onLine (repl.js:639:10)\n    at REPLServer.emit (events.js:194:15)\n    at REPLServer.EventEmitter.emit (domain.js:441:20)\n    at REPLServer.Interface._onLine (readline.js:290:10)'
```

artem-zakharchenko: I've originally included `error.message` only, but thought it would have been useful for the end users to see _their_ stack trace to the exception. I would leave error message only for now.

artem-zakharchenko: Resolved in 7342a7a.

lib/units/validateBody.js
artem-zakharchenko: Resolved in 7342a7a.

artem-zakharchenko:
@kylef, could you please review it once more?

- Adjusted the developer-friendly message to suggest to check the given JSON's validity (not the JSON Schema validity)
- Error message now includes only the parsing error (`error.message`) instead of the entire stack

lib/units/validateBody.js
```diff
@@ -148,6 +148,27 @@ function getBodyValidator(realType, expectedType) {
   return [null, validator[0]];
 }
 
+/**
+ * Parses a given JSON Schema to ensure it's a valid JSON.
+ * @param {string} schema
+ */
+const validateJsonSchema = (schema) => {
+  try {
+    JSON.parse(schema);
```
honzajavorek: I think at other places in Gavel we use `json-parse-helpfulerror` to provide even more helpful error messages than what `JSON.parse()` provides

kylef: I was looking for how the message body was done, didn't find it with `JSON.parse`.

```
src/mixins/validatable-http-message.coffee:jph = require 'json-parse-helpfulerror'
src/mixins/validatable-http-message.coffee:        jph.parse @body
src/mixins/validatable-http-message.coffee:        jph.parse @body
src/mixins/validatable-http-message.coffee:          parsed = jph.parse @expected.bodySchema
src/mixins/validatable-http-message.coffee:          jph.parse @expected.body
src/mixins/validatable-http-message.coffee:          jph.parse @expected.body
```

artem-zakharchenko: While looking over the changes yesterday I've spotted that we _already have_ a part of the body validation that attempts to parse a JSON Schema:

https://github.com/apiaryio/gavel.js/blob/bba82063cd9a7c1df1421a0764ddcd1b5ffd1cfd/lib/units/validateBody.js#L102-L109

Which means that mentioned issue scenario somehow doesn't trigger this logic. I would need to investigate. Hopefully, we don't need any new functions/checks, just need to fix the old ones.

artem-zakharchenko: So, it looks like with the current founding we must treat malformed JSON Schemas like an exception. Shall we change this to an exception then?

```js
const error = `Can't validate: expected body JSON Schema is not a parseable JSON:\n${exception.message}`; 
  
   return [error, null]; 
```

> At the moment it's being tolerated as a validation error, not stopping the process, and ending up in the public results.

artem-zakharchenko: I confirm that changing that soft handling to throwing an exception produces the same behavior:

* Stops Gavel validation to prevent operating on malformed data
* Propagates user-friendly error + JSON parsing error from `jhp`

honzajavorek: I think exception is all right. If Gavel is given schema, the user's intention is to validate according to it. If Gavel cannot parse it, then it shouldn't be silent or just warning. There's no way for Gavel to continue in that case, otherwise it could lead to false positives in validation.

PR #243, opened 2019-07-17
Core app prepare

### LINKED ISSUE #236: Current gavel.js won't work in Webpack
The jju library causes an error:

```
ERROR in ./~/jju/lib/utils.js Module not found: Error: Can't resolve 'fs' in 'apiary/node_modules/jju/lib'
```

See [start of the relevant internal discussion](https://apiary.slack.com/archives/C051PDUKE/p1562758557034700) where @kuba-kubula also suggested a solution:

> but since the code in `json-parse-helfulerror` is like 10 lines, you can get rid of the whole `jju` require, but load just `jju/parse`
> so I'd suggest throwing away `jph` and writing tiny wrapper around `jju/lib/parse` :slightly_smiling_face: which will make it a much much smaller bundle, once used as a package

Before we fix this we should introduce a test infrastructure which catches this type of problems - e.g. replacing browserify with a webpack.

artem-zakharchenko:
Prepares `gavel` to be used in the Core app.

## Changes

- Removes `json-parse-helpfulerror` dependency
- Removes `mochify` and `sinon` dev dependencies
- Removes `test:browser` pipeline
- Adds `jju` save dependency
- Adds custom `parseJson` utility wrapper around `jju.parse()` to stop relying on `fs`
- Removes ES features not supported in the stable spec (i.e. spreading in object literals)
- ~~Adds test suite that ensures `gavel` can be bundled in the browser~~
  - The build is configured to throw when any of NodeJS built-in modules are present in it. Therefore, it's the verification of browser compatibility de facto, cancelling any need of a dedicated test pipeline.
- Adds building of the `gavel` library into ~~CJS~~UMD

## GitHub

- Closes #236


lib/units/validateBody.js
```diff
@@ -1,10 +1,11 @@
-const jph = require('json-parse-helpfulerror');
+// const parseJson require('json-parse-helpfulerror');
```
honzajavorek: leftover

lib/utils/parseJson.js
```diff
@@ -0,0 +1,11 @@
+const { parse } = require('jju/lib/parse');
```
artem-zakharchenko: Only `utils.js` relies on `fs`. `parse` module should be working fine in isolation, getting rid of `fs` dependency, comparing when importing the module entirely:

```js
const jju = require('jju')
```

lib/utils/parseJson.js
```diff
@@ -0,0 +1,11 @@
...
+
+const parseJson = (json, revivew) => {
```
honzajavorek: I think this is a good idea in to have a wrapper not only because of the `fs` issue, but also because in the past it happened already that we were switching the JSON parser implementations.

kuba-kubula:
bravo

artem-zakharchenko:
CI for PR failed due to the flaky test (#139). Restart should resolve the build status.

PR #255, opened 2019-08-02
Removes "values" from field error messages

### LINKED ISSUE #254: Duplicate "values" in the result for status code
Currently if the status code doesn't match, the result looks like this:

```js
            "statusCode": {
                "valid": false,
                "kind": "text",
                "values": {
                    "expected": "242",
                    "actual": "200"
                },
                "errors": [
                    {
                        "message": "Expected status code '242', but got '200'.",
                        "values": {
                            "expected": "242",
                            "actual": "200"
                        }
                    }
                ]
            },
```

Having `values` twice there feels duplicate. Errors accompanying other fields do not seem to have `values`, they only have `message` and `location`. Is this intended or can we remove the values from the error message?

artem-zakharchenko:
## GitHub

- Fixes #254

### DIFF
--- a/lib/units/validateBody.js
+++ b/lib/units/validateBody.js
@@ -197,15 +197,13 @@ function validateBody(expected, actual) {
 
   if (realTypeError) {
     errors.push({
-      message: realTypeError,
-      values
+      message: realTypeError
     });
   }
 
   if (expectedTypeError) {
     errors.push({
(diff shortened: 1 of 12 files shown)


artem-zakharchenko:
Needs `gavel-spec` to be updated as well. Certain feature tests assert `values` in errors.

PR #256, opened 2019-08-02
Bumps major version due to removal of "errors[i].values"

artem-zakharchenko:


honzajavorek:
Second commit should be `docs:`

PR #268, opened 2019-08-13
Removes "jju", always uses JSON.parse() for expected/actual bodies

artem-zakharchenko:
## Why this change

Malformed JSON as the value of `expected` body has thrown before, since it was processed by `JsonSchema` and `SchemaV4Generator`. Those used `JSON.parse()` which throws on any JSON abnormalities.

Meanwhile, `jju.parse()` that we use for parsing of actual/body type, allows certain abnormalities in  JSON (i.e. trailing commas, single quotes). That treated the same malformed JSON as valid in `getBodyType` in `validateBody`, but threw an exception in `JsonSchema`.

## Changes

- Uses strict JSON parsing for _actual_ body only
- Uses weak (jju) JSON parsing for _expected_body


test/chai.js
```diff
@@ -45,7 +45,7 @@ to have an error at index ${currentErrorIndex} that includes property "${propNam
 ${JSON.stringify(target)}
   `,
         JSON.stringify(target),
-        JSON.stringify(expectedValue),
+        isRegExp ? expectedValue.toString() : JSON.stringify(expectedValue),
```
artem-zakharchenko: This is a minor improvement to display a RegExp in the stdout on failing tests.

test/fixtures/invalid-stringified-schema.js
```diff
@@ -1 +1,4 @@
-module.exports = `{"111!invalid string ... issue":''}`;
```
artem-zakharchenko: You won't believe, but this is actually a valid JSON per `jju`.

honzajavorek: parsed as

```
{
  "111!invalid string ... issue": ""
}
```

? that would be pretty okay, they're allowing single quotes, apparently. The first part is bordered with `""` and that's valid.

artem-zakharchenko: Yes, that was parsed into what you have written. While that's okay, it was failing the tests, since they expect this fixture to represent an _absolute monster invalid JSON_.

artem-zakharchenko:
`jju.parse()` re-creates an parsed Object by `Object.create(null)`, omitting any prototypal inheritance. The parsed objects are returned without built-in methods from the `Object` prototype. 

This throws at `amanda.validate()`, during the traversal of the Objects. Regardless, omitting a prototype makes those parsed Objects unusable by any consumer.

lib/units/validateBody.js
```diff
@@ -69,7 +69,11 @@ function getBodyType(body, contentType, httpMessageOrigin) {
   const hasJsonContentType = isJsonContentType(contentType);
 
   try {
+    // Use strict JSON parsing for actual body.
+    // This prevents any kind of normalization of trailing commas
+    // and other custom transformations applied by "jju".
```
honzajavorek: We removed jju, so this comment is extraneous now.

lib/utils/parseJson.js
```diff
@@ -1,11 +1,5 @@
-const { parse } = require('jju/lib/parse');
-
 const parseJson = (json, revivew) => {
-  try {
-    return parse(json, revivew);
-  } catch (error) {
-    throw error;
-  }
+  return JSON.parse(json, revivew);
```
honzajavorek: I'd add a comment here so it's more apparent why this is a separate module:

```js
/**
 * Parses JSON
 *
 * Previously we've been using jsonlint and jju for parsing. While we've stepped
 * back to use plain JSON.parse(), we keep our doors open to switch the parser
 * - which is the reason why this is abstracted into parseJson()
 */
const parseJSON = ...
```

package.json
```diff
@@ -15,11 +15,9 @@
     "lint": "eslint lib/**/*.js test/**/*.js",
     "test": "npm run test:server && npm run test:features",
     "test:server": "mocha \"test/**/*.test.js\"",
```
honzajavorek: I wouldn't call this a server `server`, there's no server in Gavel. It's more like `test:unit` or `test:mocha`.

test/fixtures/invalid-stringified-schema.js
honzajavorek: Ha ha

test/fixtures/invalid-stringified-schema.js
```diff
@@ -1 +1,4 @@
...
+// Note that Json Schema now uses "jju" for JSON parsing
+// which allows certain JSON abnormalities (i.e. trailing commas, single quotes).
+// Account on that to produce an invalid JSON.
```
honzajavorek: The comment can be removed now I think.

test/regression/unparseable-json.test.js
```diff
@@ -0,0 +1,100 @@
...
+      it('has error message about unparseable actual body', () => {
+        expect(result)
+          .to.have.errorAtIndex(0)
+          .withMessage(
+            /^Can't validate: actual body 'Content-Type' header is 'application\/json' but body is not a parseable JSON:/
+          );
+      });
+    });
+  });
+});
```
honzajavorek: MIssing test for unparseable JSON Schema (Gavel throws)

artem-zakharchenko: Test added in 36c9ed1.

package.json
artem-zakharchenko: Agree. Renamed in e61a532.

test/fixtures/invalid-stringified-schema.js
artem-zakharchenko: Corrected in 3aade08.

lib/utils/parseJson.js
artem-zakharchenko: Added in 3aade08.

lib/units/validateBody.js
artem-zakharchenko: Removed in 3aade08.

artem-zakharchenko:
@honzajavorek, thanks! All comments taken into account. Please, could you look again?

PR #300, opened 2019-09-25
Adds TypeScript type definitions

artem-zakharchenko:
This pull request adds a manual type declaration for Gavel.

> Note that the type definitions added in this pull request are written _manually_. They are not generated from the code base and, therefore, will not update if the source code changes. 

## Motivation

For Gavel to be usable within TypeScript environment.

### DIFF
--- a/typings.d.ts
+++ b/typings.d.ts
@@ -1,0 +1,80 @@
+declare module 'gavel' {
+  export enum RESTMethods {
+    GET = 'GET',
+    POST = 'POST',
+    PUT = 'PUT',
+    PATCH = 'PATCH',
+    DELETE = 'DELETE',
+    OPTIONS = 'OPTIONS'
+  }
+
+  export interface HTTPMessage {
+    method: RESTMethods;
+    uri: string;
+    statusCode: string | number;
+    headers: Record<string, string> | string;
+    body: Record<string, any> | string;
+    /**
+     * [JSON Schema](https://json-schema.org/) Draft V3-4.
+     */
+    bodySchema: Record<string, any> | string;
+  }
+
+  export enum FieldKind {
+    text = 'text',
+    json = 'json'
+  }
+
+  export interface ValidationResult {
+    /**
+     * Indicates whether the actual HTTP message is valid
+     * against the expected HTTP message.
+     */
+    valid: boolean;
+    /**
+     * Validation results of each individual HTTP message
+     * field (i.e. "statusCode", "body", etc).
+     */
+    fields: Record<string, FieldValidationResult>;
+  }
+
+  export interface FieldValidationResult {
+    /**
+     * Indicates whether a single HTTP message field is valid.
+     */
+    valid: boolean;
+    /**
+     * Kind of validation that has been applied to the field.
+     */
+    kind: FieldKind | null;
+    /**
+     * Normalized HTTP message field values that are being validated.
+     */
+    values: {
+      expected: any;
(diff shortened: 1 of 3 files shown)


artem-zakharchenko:
I've tested the typings by linking the Gavel package and using it in a TypeScript project. I confirm that the typings are used and they are valid.

typings.d.ts
```diff
@@ -0,0 +1,80 @@
...
+  export interface HTTPMessage {
+    method: RESTMethods;
+    uri: string;
+    statusCode: string | number;
+    headers: Record<string, string> | string;
+    body: Record<string, any> | string;
+    /**
+     * [JSON Schema](https://json-schema.org/) Draft V3-4.
+     */
+    bodySchema: Record<string, any> | string;
```
kylef: This is more of a question about TypeScript than PR comment.

Context: In the future JSON Schema Draft 6, booleans will also valid schema anywhere. Whereas in Draft 4 they are only permittable in 'additionalProperties' and 'additionalItems' properties. [(true schema validates like a `{}` schema, false schema validates like `{"not": {}}`)](https://github.com/kylef/JSONSchema.swift/blob/302a7dbb4f3a85cdb8afa5632b7058572a60eea0/Sources/Draft4Validator.swift#L48). 

Question: Will adapting this type definition to support JSON Schema Draft 6 become a breaking change to gavel.js as this is part of the interface/contract to library consumers. Which would cause major version of gavel.js?

kylef: It would be neat if `Record<string, any>` would be type definition for JSON Schema (Draft X) itself.

artem-zakharchenko: I would put this question other way: will we deprecate JSON Schema Draft > 5 when introducing Draft V6+?

If not, the proper way would be to use a union type:

```ts
interface HTTPMessage {
  bodySchema: JSONSchemaV6 | JSONSchemaV5 | JSONSchemaV4
}
```

Types derive from the implementation, so if Gavel supports all those JSON Schema draft versions, then there is no breaking change. If we decide to deprecate older versions support, that would be a breaking change by definition, since the actual code of the library will change, not only types.

artem-zakharchenko:
Thanks for the review!

> * Is it possible to lint or validate the typing file?
> * Could there be a CI step to ensure that the file is atleast valid type script?
> * How can we ensure that changes to Gavel won't break the typings?

All these points are valid and are solved by generating type definitions from the source code. I'd absolutely love to do that, but it's a much bigger task which I see no need to solve at the moment. Any kind of manual linting would be a re-inventing of bicycle.

I'm writing this task down (https://github.com/apiaryio/gavel.js/issues/301) so it's evident and not forgotten.

typings.d.ts
artem-zakharchenko: Let me see if there is any typings exposed by JSONSchema in any way, so we could inherit it here. That'd be great.

artem-zakharchenko: It turns out there is `@types/json-schema`, which provides type definitions for JSONSchema Draft V4, V6, and V7. We can include the V4 in Gavel type definitions.

artem-zakharchenko: Although `@types/json-schema` provides type definition for JSON Schema Draft V4, since we don't have the same for V3, we would need to use a union type with `Record`, making any more precise typings useless. TypeScript won't validate the validity of JSON Schema V4 since, if invalid, it satisfies the `Record` type.

PR #355, opened 2019-11-26
Adds JSON Schema Draft 6/7 support

### LINKED ISSUE #90: Use ajv JSON Schema Validator
Suggestion from https://github.com/apiaryio/dredd/issues/278.

artem-zakharchenko:
Contains backward-compatible changes that enable JSON Schema Draft 6/7 support (pre-requisite for OpenAPI support in Dredd).

## Changelog

- Splits legacy (Amanda, TV4) and "future" AJV validators into separate classes
- Provides a master-class to determine a proper validator based on the JSON Schema explicit or implicit version
- Adjusts existing unit/integration test suites to align with the introduced changes
- **FIX:** Makes `errors[n].location.property` names consistent and validator-agnostic (`property` is not always an array that points to the deeply nested property, i.e. `['friends', 2, 'name']`)
- Added a single point of coercion of arbitrary validation structure to Gavel-compliant validation structure (`toGavelResults()`) and respective unit tests

## GitHub

- Closes #90 

## Roadmap

- [x] Split legacy and next JSON Schema Draft versions validation into separate validator classes
- [x] Provide a master class that would determine which validator class to use
- [x] Resolve "schema with key or id already exists" during V6-7 validation with AJV
- [x] Verify new schemas V6-7 are valid representations of their draft versions
- [x] Discuss/Revert previously present implicit logic to waterfall attempts to validate a given JSON Schema without an explicit `$schema` version (first as V3, then as V4)
- [x] Revert support of implicit JSON Schema version (fallback to V3 if no version is set via `$schema` property)
- [x] Adjust derived `JsonExample` class to implement `JsonSchemaValidator` properly
- [x] Adjust the usage surface of `new JsonSchema()`
  - [x] validateBody
  - [x] validateHeaders
- [x] Adjust affected tests
  - [x] `test/unit/validators/json-schema.test`
  - [x] `amanda-to-gavel-shared`
  - [x] `TextDiff` unit tests
  - [x] `validate` integration tests
- [x] Add new tests
  - [x] Unit test to assert that amanda/tv4/ajv error format is coerced to Gavel-compliant format correctly
- [x] Infer a JSON Schema Draft V7 if a given JSON Schema contains a single boolean value


lib/errors.js
```diff
@@ -7,6 +7,7 @@ class UnknownValidatorError extends Error {}
 class NotValidatableError extends Error {}
 class NotEnoughDataError extends Error {}
 class JsonSchemaNotValid extends Error {}
+class JsonSchemaNotSupported extends Error {}
```
artem-zakharchenko: Explicit error type to use in unit tests instead of asserting an exact message.

lib/validators/json-schema-legacy.js
```diff
@@ -0,0 +1,182 @@
...
+    divisibleBy: (prop, val, validator) =>
+      `The ${prop} property is not divisible by ${validator} (current value is ${JSON.stringify(
+        val
+      )}).`,
+    uniqueItems: (prop) => `All items in the ${prop} property must be unique.`
+  }
+};
+
+class JsonSchemaLegacy extends JsonSchemaValidator {
+  validateSchema() {
```
artem-zakharchenko: Overriding `JsonSchemaValidator.validateSchema()` to use TV4 to validate both V3 and V4 (copying previous implementation, refactored).

lib/validators/json-schema.js
```diff
@@ -1,321 +1,17 @@
...
-      `The ${prop} property is not divisible by ${validator} (current value is ${JSON.stringify(
-        val
-      )}).`,
-    uniqueItems: (prop) => `All items in the ${prop} property must be unique.`
-  }
-};
+const { JsonSchemaLegacy } = require('./json-schema-legacy');
+const { JsonSchemaValidator, getSchemaVersion } = require('./json-schema-next');
 
 class JsonSchema {
```
artem-zakharchenko: A wrapping class to determine which actual validator class to use. A temporary wrapper until we deprecate Amanda and TV4. Overkill: determines a JSON Schema's version _twice_ (once in the wrapper class, another time in the actual validator class).

artem-zakharchenko:
@kylef, please, do you mind giving my approach a look? I would like to verify my vision is optimal before I proceed. Thanks.

test/unit/validators/json-schema.test.js
```diff
@@ -71,7 +71,7 @@ describe('JsonSchema', () => {
               assert.isTrue(validator.errors instanceof ValidationErrors);
             });
 
-            it('should return some errors', () => {
+            it.only('should return some errors', () => {
```
kylef: Oops

artem-zakharchenko: There can be an eslint rule that prevents such commits. I'd suggest to adopt that to prevent such situations.

artem-zakharchenko:
I've removed `amanda-to-gavel-shared` module entirely, since I've also removed `evaluateOutputToResults()`. Instead, `JsonSchemaLegacy` would use an internal method to coerce Amanda/TV4 result format to Gavel-compliant error format.

I will add a new test suite to cover that the newly added method converts Amanda/TV4/AJV result format to Gavel-compliant error format properly as the next step.

test/unit/validators/json-schema.test.js
artem-zakharchenko: Resolved.

test/unit/utils/to-gavel-result.test.js
```diff
@@ -0,0 +1,87 @@
...
+        length: 2,
+        0: {
+          property: ['users', 'username'],
+          propertyValue: 123,
+          attributeName: 'type',
+          attributeValue: 'string',
+          message: 'Amanda error message'
+        },
+        1: {
+          property: ['friends[2]', 'online'],
```
artem-zakharchenko: Amanda produces `prop[n]` property, which appears to be invalid JSON pointer.

I've refactored the coercing logic so it produces `['fiends', 2, 'online']` JSON pointer format.

artem-zakharchenko:
@kylef, could you please take a look at this pull request? It should be ready. Thanks 😊

test/unit/utils/to-gavel-result.test.js
artem-zakharchenko: This would be considered a breaking change, but it's rather a bug fix, because it fixes an unexpected property format. It's expected to be a list of stings, and it was resolved to a plain string sometimes.

lib/validators/json-schema-next.js
```diff
@@ -0,0 +1,173 @@
...
+    return validationResult.valid;
+  }) || [null, null];
+
+  return schemaVersion;
+};
+
+class JsonSchemaValidator {
+  constructor(jsonSchema) {
+    this.jsonSchema = jsonSchema;
+    this.jsonSchemaVersion =
```
kylef: In Draft 6 and onwards, `true` and `false` are valid schemas (https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.4.3.2). I think it would be worth testing these two values in regards to the version detection. I'd gather it would end up being draft 3/4 as implicit schema and then become an invalid schema.

kylef: Perhaps we would want to default it to Draft 7 under those circumstances since the behaviour would be same in draft 6, 7 and 2019-09

artem-zakharchenko: Inferring of JSON Schema Draft 7 implemented in https://github.com/apiaryio/gavel.js/pull/355/commits/593ea41dc86af6b935f7e45645a8bf13f728944f.

PR #366, opened 2019-12-12
Uses "terser" plugin to compress the built bundle

artem-zakharchenko:
The current file size of the built bundle is `508KB`. That is caused by a number of factors:

1. The bundle is not compressed.
1. The bundle includes `ajv` module, which size is roughly `200KB`.

This pull request enables bundle compression using `rollup-plugin-terser`.

## Bundle size analyzis

Here's the analyzis of the uncompressed bundle:

Screenshot 2019-12-12 at 14 19 46

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -750,6 +750,12 @@
       "integrity": "sha1-M3dm2hWAEhD92VbCLpxokaudAzc=",
       "dev": true
     },
+    "buffer-from": {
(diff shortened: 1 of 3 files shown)


artem-zakharchenko:
We can also use CI tools like [bundlesize](https://github.com/siddharthkp/bundlesize) to monitor bundle size deviations. It's proven to be useful.

PR #368, opened 2019-12-13
Adds deprecation warning on JSON Schema Draft 3 usage

### LINKED ISSUE #361: Deprecate JSON Schema Draft 3 (Amanda) support
## What

I suggest to deprecate JSON Schema Draft 3 support in Gavel (consequently in Dredd).

These changes must be brought in two phases described below.

### Stage 1

1. Usage of JSON Schema Draft 3 is met with a deprecation warning that a) states the end of support for Draft 3 schemas, b) suggests an alternative (v4?)
1. JSON Schema Draft 3 are not longer used as an inferred schema version in case no explicit version is provided. Instead, Draft 4 is used as the default implicit JSON Schema version.

### Stage 2

1. JSON Schema Draft 3 is removed from the library's internals.
1. `amanda` is uninstalled.
1. Codebase is adjusted according to the removed Draft 3 support.

artem-zakharchenko:
This pull request adds a deprecation warning whenever attempted to validate JSON Schema Draft V3 (Amanda). JSON Schema Draft 3 support is going to be dropped in the next major release of Gavel.

## GitHub

- Closes #361

### DIFF
--- a/lib/utils/deprecated.js
+++ b/lib/utils/deprecated.js
@@ -6,10 +6,9 @@
  * @param {string} message
  */
 module.exports = function deprecated(message) {
-  const isNode = typeof process !== 'undefined';
-  const shouldSuppressWarnings =
-    isNode && process.env.SUPPRESS_DEPRECATION_WARNINGS;
-  const shouldOutputMessage = isNode ? !shouldSuppressWarnings : true;
+  const shouldOutputMessage =
+    typeof process === 'undefined' ||
+    !process.env.SUPPRESS_DEPRECATION_WARNINGS;
 
   if (shouldOutputMessage) {
     console.warn(`DEPRECATED: ${message}`);
--- a/lib/utils/deprecated.js
+++ b/lib/utils/deprecated.js
@@ -1,0 +1,17 @@
+/* eslint-disable no-console */
+
+/**
+ * Outputs a deprecation message.
+ * Supports silencing the messages based on the environmental variable.
+ * @param {string} message
+ */
+module.exports = function deprecated(message) {
+  const isNode = typeof process !== 'undefined';
+  const shouldSuppressWarnings =
+    isNode && process.env.SUPPRESS_DEPRECATION_WARNINGS;
+  const shouldOutputMessage = isNode ? !shouldSuppressWarnings : true;
+
+  if (shouldOutputMessage) {
+    console.warn(`DEPRECATED: ${message}`);
+  }
+};
--- a/lib/validators/json-schema-legacy.js
+++ b/lib/validators/json-schema-legacy.js
@@ -5,7 +5,7 @@ const jsonPointer = require('json-pointer');
 const { JsonSchemaValidator, META_SCHEMA } = require('./json-schema-next');
 const { ValidationErrors } = require('./validation-errors');
 const toGavelResult = require('../utils/to-gavel-result');
-const deprecated = require('../utils/deprecated');
+const warnOnDeprecation = require('../utils/warn-on-deprecation');
 
 /**
  * Returns a proper article for a given string.
@@ -77,7 +77,7 @@ class JsonSchemaLegacy extends JsonSchemaValidator {
 
     switch (this.jsonSchemaVersion) {
       case 'draftV3':
-        deprecated(
+        warnOnDeprecation(
           'JSON Schema Draft V3 is deprecated. Please use a newer version of JSON Schema (Draft V4-7).'
         );
 
--- a/lib/validators/json-schema-legacy.js
+++ b/lib/validators/json-schema-legacy.js
@@ -5,6 +5,7 @@ const jsonPointer = require('json-pointer');
 const { JsonSchemaValidator, META_SCHEMA } = require('./json-schema-next');
 const { ValidationErrors } = require('./validation-errors');
 const toGavelResult = require('../utils/to-gavel-result');
+const deprecated = require('../utils/deprecated');
 
 /**
  * Returns a proper article for a given string.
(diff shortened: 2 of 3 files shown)


artem-zakharchenko:
Updated the deprecation message to be shorter, straight to the point and mentions exact JSON Schema Draft versions in migration suggestion.

artem-zakharchenko:
@kylef, could you please look into this? If we decide to deprecate JSON Schema Draft V3 as a breaking change I'd like to merge this first, so we have a proper deprecation lifecycle. Thanks.

lib/validators/json-schema-legacy.js
```diff
@@ -76,6 +76,10 @@ class JsonSchemaLegacy extends JsonSchemaValidator {
 
     switch (this.jsonSchemaVersion) {
       case 'draftV3':
+        // eslint-disable-line no-console
+        console.warn(
```
kylef: I'm not sure that stdout is perhaps the right place for this. Although, I don't have any other suggestion.

Depending on how you use the library:

- On web service - it can flood the logs with this deprecated message, and it may be in the wrong format (since our services used structured log formats in JSON).
- In the web - the message is hidden in the console and usually user's are not going to see it
- Dredd -- this is one case it may actually expose the message, although it might get hidden by the test output.

artem-zakharchenko: You can see this warning only in stdout and browser's console. You are right that it's not intended for indirect consumers (Apiary or Dredd users). Warning those should be handled by respective parent environment (i.e. warning message in the UI of Apiary). I don't think we should (and can) handle this on Gavel's side. 

What I think we should do is:

1. Determine amount of users using Draft V3 in production.
1. See if there is a substantial difference between drafts V3 and V4 (as you've mentioned there doesn't seem to be any).
1. If any of the preceding points give us reason to do any UI changes then we should, otherwise there is no changes that users should do.

patricksmms: I'm wondering if it would make sense to put it on the validation result either inside `errors` or a new property `warnings`. But I don't know how these are consumed.

artem-zakharchenko: @patricksmms it's a perfectly fine suggestion, no worries. 

When we re-designed the output of Gavel we put certain restrictions over what should be present in the validation result and `errors` in particular. Shortly, you should access those properties only if any validation has happened. Also, validation result always represents the result of matching actual HTTP message against the expected one.

Keeping that in mind, such warning is not a result of expected/actual data validation. Instead, it's a result of validating any of that data against an external factor (internally supported version of JSON Schema). We shouldn't have such messages in the validation result by design.

> It's also useful to know that Gavel provides no validation against external factors, for example that the given HTTP message is compliant with its respective specification. I'd like for it to stay that way.

artem-zakharchenko: @kylef if preventing spamming the logs is our primary concern we make Gavel accept options. One of those options may be `suppressDeprecationWarnings: true`, which we can pass to Gavel calls on the Apiary side. Using Gavel directly, or under Dredd should still produce a warning evident for the developer (in stdout).

artem-zakharchenko: **Okay, so to recap:**

- Gavel can be consumed directly (via its NodeJS API) and indirectly (Dredd, Apiary). 
- Deprecation message should be evident, so it's very easy to spot for the user.

Correspondingly, deprecations should be handled for both direct and indirect Gavel consumers.

## Direct consumer

I'm not aware of a better way than printing a deprecation message into `stdout`. It's also how other libraries handle deprecations, as far as I've seen.

Since Gavel can run in NodeJS indirectly, it _may_ pollute a parent's process stdout. 

## Indirect consumer

 Indirect consumer should know that some deprecation message was thrown, and propagate it in some form to the actual user. Displaying a message banner in the UI is a nice example for the client-side consumption.

patricksmms: This makes sense to me.

artem-zakharchenko:
I've added the `deprecated` utility that does the following:

1. Checks if the code is executed in NodeJS.
1. If yes, checks for the environmental flag that would silence the deprecations (`SUPRESS_DEPRECATION_WARNINGS` as of now). I think such flag should be absurdly lengthy to indicate you shouldn't be using it on a regular basis.
1. If no, the code runs in the browser and deprecation warnings must be shown.

lib/utils/deprecated.js
```diff
@@ -0,0 +1,17 @@
...
+ */
+module.exports = function deprecated(message) {
+  const isNode = typeof process !== 'undefined';
+  const shouldSuppressWarnings =
+    isNode && process.env.SUPPRESS_DEPRECATION_WARNINGS;
+  const shouldOutputMessage = isNode ? !shouldSuppressWarnings : true;
+
+  if (shouldOutputMessage) {
+    console.warn(`DEPRECATED: ${message}`);
+  }
```
patricksmms: Here the `isNode` check is repeated unnecessarily.
I would simplify all this to:
```js
    const shouldOutputDeprecationWarnings = process === 'undefined' ? true : !process.env.SUPPRESS_DEPRECATION_WARNINGS;

    if (shouldOutputDeprecationWarnings) {
      console.warn(`DEPRECATED: ${message}`);
    }
```
Eventually extracting `process === 'undefined'` to `isNode`.

artem-zakharchenko: Good one, thanks! Will adjust.

artem-zakharchenko:
Strangely, Travis build for PR is pending since the yesterday's noon.

lib/utils/deprecated.js
artem-zakharchenko: I would even settle on this:

```js
const shouldOutputMessage =
  typeof process === 'undefined' ||
  !process.env.SUPPRESS_DEPRECATION_WARNINGS;
```

lib/validators/json-schema-legacy.js
```diff
@@ -76,6 +77,10 @@ class JsonSchemaLegacy extends JsonSchemaValidator {
 
     switch (this.jsonSchemaVersion) {
       case 'draftV3':
+        deprecated(
```
patricksmms: Somehow the comment on this line got lost. I suggest giving a name that conveys the action of the function, so you don't need to read the function to understand what it does:
 ```suggestion
        logDeprecatedWarning(
```

artem-zakharchenko: Renamed to `warnOnDeprecation`. Thanks.

PR #369, opened 2019-12-13
Removes JSON Schema Draft 3 support

artem-zakharchenko:
> **This pull request introduces a breaking change.** It's a successor to #368.

- Removes `amanda` dependency
- Adjusts `json-schema-legacy` to not handle any draftV3 schemas
- Treat `draftV3` as unsupported, throw an exception
- Removes Amanda-related tests
- **Makes JSON Schema Draft V4 the default (fallback) schema version when no explicit version is set**

## Roadmap

- [x] Update Gavel spec (https://github.com/apiaryio/gavel-spec/pull/67)


artem-zakharchenko:
I've marked the `chore` commit with a "BREAKING CHANGE" commit body.

PR #404, opened 2020-01-16
Uses AJV for JSON Schema Draft V4 validation

artem-zakharchenko:
## Changes

- Draft 4 is no longer considered legacy (in terms of which JSON Schema validator is being used)
- Uses AJV to validate JSD4 instead of TV4
- Preserves TV4 for implicit JSON Schema generation (from a body literal and headers)
- Adjusts TV4 Schema Generator to be a valid representation of JSD4 (`id` -> `$id`)

## Issues

- **Requires #369**
- Relates to #90 
- Relates to https://github.com/apiaryio/dredd/issues/1647


lib/validators/headers-json-example.js
```diff
@@ -1,5 +1,5 @@
 const errors = require('../errors');
-const { JsonSchemaLegacy } = require('./json-schema-legacy');
+const { JsonSchemaValidator } = require('./json-schema-next');
```
artem-zakharchenko: Since Header validation is based on an implicitly generated JSON Schema Draft 4, we cannot use the legacy class anymore. The legacy class is left only to support Draft 3.

artem-zakharchenko:
I'd like to merge this after Amanda is fully deprecated in https://github.com/apiaryio/gavel.js/pull/369. That way we could get rid of JsonSchemaLegacy class completely.

artem-zakharchenko:
I've added a regression test for JSON Schema Draft 4 usage of `allOf` and `additionalProperties` (originated in https://github.com/apiaryio/dredd/issues/1647). Related to the changes since AJV is required to properly validated such situation for JSD4.

artem-zakharchenko:
I've postponed the introduction of the `schemaPath` property to `errors[i].location`. The reason is that it would require to be added to the spec explicitly (feature tests assert the exact shape of the  `location` object). I have worries about this making `location` to exact in the spec, while it should be quite arbitrary by design. I think we don't have this properly introduced yet.

artem-zakharchenko:
Rebased upon the latest `master` to include the removal of Amanda. 

- Removed `JsonSchema` and `JsonSchemaLegacy` classes
- Adjusted affected validators and tests

lib/validators/json-schema-next.js
```diff
@@ -90,6 +95,28 @@ class JsonSchemaValidator {
     }
   }
 
+  /**
+   * Resolve stringified JSON Schema to an Object.
```
kylef: I wonder if this is the right terminology, as a JSON Schema does not have to be an object, just valid JSON as boolean schemas are permitted.

PR #536, opened 2020-11-27
Build CommonJS version of gavel

realityking:
This PR adds a CommonJS version of this module for the benefit of Node use-cases including `dredd`.

TBH I'm not sure why this package is build as an UMD module. I suspect this might be how it's deployed for Apiary UI? In that case this might require some change as the filename of the UMD build changes with this PR.

Not only is CommonJS code generating less closures than UMD, it also doesn't bundle all dependencies. This allows `dredd` to load, for example, `media-typer` only once.

I also moved `url` to be a dev dependency as it's only used to build the UMD version. For Node.js the built-in version of `url` is preferred.


realityking:
Revisiting this PR, it doesn't seem like `gavel` is build with browsers in mind at all. Is there a reason to ship the UMD version at all? It might be good to only ship the CommonJS build.

PR #535, opened 2020-11-27
Upgrade ajv to version 6.12.6

realityking:
Just a small upgrade to aid in de-duplication

PR #565, opened 2021-03-19
Update ajv, deep-equal, json-pointer, commander and other dependencies

kuba-kubula:
closes #252
closes #481

closes #520
closes #535
closes #539
closes #540

closes #547

closes #558

closes #562

closes #564


package.json
```diff
@@ -30,45 +30,44 @@
...
     "curl-trace-parser": "0.0.10",
-    "deep-equal": "1.1.1",
+    "deep-equal": "2.0.5",
     "http-string-parser": "0.0.6",
-    "json-pointer": "0.6.0",
+    "json-pointer": "0.6.1",
     "media-typer": "1.1.0",
     "tv4": "1.3.0",
-    "url": "0.11.0"
+    "url-parse": "1.5.1"
```
realityking, line 48: It'd be great if you could have a look at #536. Technically `url` is just a dev-dependency right now.

kuba-kubula: Based on the code of validate.js - there's still the `validateURI` function using deep-clone to do comparison of a query-string regardless of ordering of `key=value` pairs.

kuba-kubula: building CommonJS sure - I might look into it.
You're right that `url` could be a devDependency, as NodeJs has its version as part of core.
Have to 🤔 about it a bit more.

kuba-kubula: @realityking I'm more or less thinking about actually reusing `uri-js` instead of any url-parse/url/native-url or what-not (that package is anyway used by `ajv`, so it is present at all cost)

package.json
```diff
@@ -30,45 +30,44 @@
...
-    "tv4": "1.3.0",
-    "url": "0.11.0"
+    "tv4": "1.3.0"
   },
   "devDependencies": {
-    "@rollup/plugin-json": "4.0.1",
-    "chai": "4.2.0",
-    "cross-spawn": "7.0.1",
+    "@rollup/plugin-json": "4.1.0",
+    "caseless": "0.12.0",
```
artem-zakharchenko, line 48: Why are we moving `caseless` into `devDependencies`? It's used on runtime:

https://github.com/apiaryio/gavel.js/blob/22ccd2ddb659fe252cff55ea3de74938db557d5a/lib/utils/tv4-to-headers-message.js#L14

The same is true for the `json-pointer`:

https://github.com/apiaryio/gavel.js/blob/22ccd2ddb659fe252cff55ea3de74938db557d5a/lib/utils/to-gavel-result.js#L19

That's also the reason the CI fails at the moment.

artem-zakharchenko (changes requested):
The `url` update is good, but I've left some comments about moving some dependencies into dev dependencies.

package.json
realityking: There's a related discussion in https://github.com/apiaryio/gavel.js/pull/537

package.json
realityking: Not a bad idea. Another idea could be to move to the native URL API (https://nodejs.org/api/url.html#url_the_whatwg_url_api) and either polyfill for older browsers or ask users to polyfill themselves.

package.json
kuba-kubula: @artem-zakharchenko both these packages are relevant only to tests:
- file `test/unit/utils/tv4-to-headers-message.test.js` for `caseless`
- file `test/unit/utils/to-gavel-result.test.js` for `json-pointer`

both can be moved into devDeps.
I'll move these utility files into test folder and that should solve the failing `lint` part of tests, ok?

artem-zakharchenko (approved):

